PR #7, opened 2024-01-16
Remove a11y test as it is for the LA doc upload flow

vrajmohan:


.github/workflows/run-tests.yaml
```diff
@@ -61,46 +61,3 @@ jobs:
...
-    runs-on: ubuntu-latest
-    permissions:
-      actions: read
-    services:
-      postgres:
-        image: postgres
-        env:
-          POSTGRES_USER: mdbenefits_test
-          POSTGRES_DB: mdbenefits_test
-          POSTGRES_PASSWORD: postgres
```
bseeger, line 75: Do we want to take these out or just fix them?

bseeger: Never mind. We can add this logic back in if we need to. The tests really do need to go as they don't apply to our setup at all.

bseeger (approved):
LGTM!

PR #13, opened 2024-01-18
Disable schema creation using JPA

vrajmohan:
Schema creation should be left to Flyway. This is a familiar problem that was also encountered in the form-flow library.

### DIFF
--- a/src/main/resources/application-test.yaml
+++ b/src/main/resources/application-test.yaml
@@ -25,9 +25,6 @@ spring:
     clean-disabled: false
   thymeleaf:
     prefix: classpath:/templates/
-  jpa:
-    hibernate:
-      ddl-auto: create-drop
   main:
     allow-circular-references: true
   session:


src/main/resources/application-test.yaml
```diff
@@ -25,9 +25,6 @@ spring:
     clean-disabled: false
   thymeleaf:
     prefix: classpath:/templates/
-  jpa:
-    hibernate:
-      ddl-auto: create-drop
```
bseeger, line 30: I thought this just dropped the tables at the end of testing.  Is there something that cleans up data from tests?

bseeger: That being said, I'm not super clear on how this config affects things.

vrajmohan: See https://docs.spring.io/spring-boot/docs/1.1.0.M1/reference/html/howto-database-initialization.html#howto-initialize-a-database-using-hibernate. This was the root cause of https://www.pivotaltracker.com/story/show/186402340.

PR #15, opened 2024-01-19
Update parish to county

bseeger:
* changes parish to county


src/main/java/org/mdbenefits/app/data/enums/Counties.java
```diff
@@ -0,0 +1,38 @@
+package org.mdbenefits.app.data.enums;
+
+public enum Counties {
+  ALLEGANY("Allegany"),
```
vrajmohan, line 4: Minor - I would think (this is a question for Programs) that they would want the suffix County for all except Baltimore City:
1. Due to the fact that Baltimore City and Baltimore County are _both_ counties
2. Due to the fact that the counties with a possessive name sound weird without the suffix - e.g., Prince George's

src/main/java/org/mdbenefits/app/utils/SubmissionUtilities.java
```diff
@@ -89,7 +90,7 @@ public static String formatMoney(Double value) {
     }
 
     public static boolean isNolaParish(Submission submission) {
-        return Parish.ORLEANS.name().equals(submission.getInputData().get("parish"));
```
vrajmohan, line 92: Can't we get rid of the calls to this method easily?

vrajmohan (approved):
Minor comments

src/main/java/org/mdbenefits/app/data/enums/Counties.java
```diff
@@ -0,0 +1,38 @@
...
+  ANNE_ARUNDEL("Anne Arundel"),
+  BALTIMORE_CITY("Baltimore City"),
+  BALITMORE("Baltimore"),
```
bseeger, line 7: Oh! Typo! I'll fix this soon.

src/main/java/org/mdbenefits/app/data/enums/Counties.java
bseeger: Yeah, I thought about this as well. The only annoying part is if we have to remove the word "County" when filling out forms.  Let's see what Programs thinks.

bseeger: I can add "County" in and then we can take it out, if need be.

src/main/java/org/mdbenefits/app/utils/SubmissionUtilities.java
bseeger: I'll see if it opens up a rabbit hole. If not, I'll take it and calls to it out.

src/main/java/org/mdbenefits/app/data/enums/Counties.java
vrajmohan: IIRC, this doesn't go on a form.

PR #16, opened 2024-01-19
Remove Spanish and Vietnamese message resources

vrajmohan:
Without this change, you could confuse yourself by hitting the endpoint with a lang parameter.


bseeger (approved):
Looks good!

PR #24, opened 2024-01-23
Change accepted file types based on feedback

bseeger:
This also makes the message for what file types are acceptable in the information reveal dynamic.  It was a static message that now will not need to be changed if the file types, etc change again.

Based on feedback on this ticket: https://www.pivotaltracker.com/story/show/186854106


vrajmohan (approved):
Looks good. Should we create a story in the backlog to add back the image types?

bseeger:
Good point. I added this ticket: https://www.pivotaltracker.com/story/show/186916542

PR #25, opened 2024-01-23
Change the icons to use an orange background

bseeger:


vrajmohan (approved):
LGTM. Can we get rid of the nola-* HTML?

vrajmohan:
👍🏾

PR #27, opened 2024-01-24
Make doc uploader use an accordion versus reveal

bseeger:
This also overrides the honeycrisp version of the accordion so that we can pass it an html item versus just text.

This should complete https://www.pivotaltracker.com/story/show/186854106

### DIFF
--- a/src/main/resources/templates/fragments/docUploadSeeWhatToAddReveal.html
+++ b/src/main/resources/templates/fragments/docUploadSeeWhatToAddReveal.html
@@ -1,7 +1,7 @@
 <th:block th:fragment="docUploadSeeWhatToAddReveal">
   <th:block th:replace="~{fragments/accordion :: accordion(
     buttonLabel=~{::contentLabel},
-    content=~{::contentContent})}">
+    content=~{::content})}">
     <th:block th:ref="contentLabel">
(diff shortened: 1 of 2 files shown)


src/main/resources/templates/fragments/docUploadSeeWhatToAddReveal.html
```diff
@@ -1,15 +1,14 @@
 <th:block th:fragment="docUploadSeeWhatToAddReveal">
-  <th:block th:replace="~{'fragments/honeycrisp/reveal' :: reveal(
-                  controlId='r1',
-                  linkLabel=~{::revealLabel1},
-                  content=~{::revealContent1})}">
-    <th:block th:ref="revealLabel1">
+  <th:block th:replace="~{fragments/accordion :: accordion(
+    buttonLabel=~{::contentLabel},
+    content=~{::contentContent})}">
```
vrajmohan, line 4: The name :-(

bseeger: @vrajmohan  just updated this to just be "content".  I had been wondering about naming :)

vrajmohan (approved):
LGTM

PR #29, opened 2024-01-29
Remove language preference as the pilot app is for English applicants only

vrajmohan:


bseeger (approved):
LGTM! The work I'm doing on the starting screens overlaps a little, but didn't clean up language preferences fully.

PR #32, opened 2024-02-01
Update Sign and Submit screens

vrajmohan:


src/main/java/org/mdbenefits/app/submission/actions/SetDefaultSensitiveConvictionAnswers.java
```diff
@@ -0,0 +1,32 @@
...
+            "IsViolatingParole",
+            "ConvictedForLyingAboutBenefits",
+            "ConvictedForTradingBenefits",
+            "IsReceivingBenefitsWithFakeID"
+    };
+
+    @Override
+    public void run(Submission submission) {
+        Arrays.stream(SENSITIVE_CONVICTION_QUESTIONS).forEach(s -> submission.getInputData().putIfAbsent("noOne" + s + "[]", "true"));
+    }
```
bseeger, line 31: Why are you creating a list like name here when the value isn't a list?  Maybe you don't need the "[]"?

src/main/java/org/mdbenefits/app/preparers/ApplicantDetailsPreparer.java
```diff
@@ -36,6 +39,20 @@ public Map<String, SubmissionField> prepareSubmissionFields(Submission submissio
         results.put("applicantSSN", new SingleField("applicantSSN", SubmissionUtilities.formatSSN((String) inputData.get("encryptedSSN")), null));
         results.put("speaksEnglish", new SingleField("speaksEnglish", (String) "Yes", null));
 
+        prepareAnswersToSensitiveQuestions(inputData, results);
         return results;
     }
+
+    private static void prepareAnswersToSensitiveQuestions(Map<String, Object> inputData, Map<String, SubmissionField> results) {
+        for (String question : SetDefaultSensitiveConvictionAnswers.SENSITIVE_CONVICTION_QUESTIONS) {
```
bseeger, line 47: Pulling these from a specific Action feels weird, but maybe having as an enum feels weird as well.

src/main/java/org/mdbenefits/app/submission/actions/SetDefaultSensitiveConvictionAnswers.java
bseeger: You'd only use the `[]` if the original var looked like this: 
`private List<String> noOneHasDrugKingpinFelony;`

Or at least that's what we've used the brackets to indicate so far.

src/main/java/org/mdbenefits/app/submission/actions/SetDefaultSensitiveConvictionAnswers.java
```diff
@@ -0,0 +1,32 @@
...
+}
```
bseeger, line 32: Adding a test for this class would be useful.

src/main/java/org/mdbenefits/app/preparers/ApplicantDetailsPreparer.java
vrajmohan: Agreed, I couldn't think of a proper place to put it.

src/main/java/org/mdbenefits/app/submission/actions/SetDefaultSensitiveConvictionAnswers.java
bseeger: We talked through this and I understand more now.  Resolving this.

bseeger (approved):
LGTM - thanks for the changes.

PR #33, opened 2024-02-05
Updates based on feedback for document uploader

bseeger:
In /src/test, this moves "utils" -> "testutils" to make a space to do unit tests for main "utils".

### DIFF
--- a/src/main/resources/messages.properties
+++ b/src/main/resources/messages.properties
@@ -1159,7 +1159,7 @@ add-document-types.doc-type-label.correspondence.other=Other
 # doc upload intro
 doc-upload-intro.title=Doc upload intro
 doc-upload-intro.header=Would you like to add documents?
-doc-upload-intro.subtext=To get benefits, you will need to give documents that show things like how much money you get, costs to live and where you live.<br/><br/>Adding them now can help the process go faster. <br/> <br/><b>If you don't know what to add, skip it and you can talk with a county worker later.</b>
+doc-upload-intro.subtext=To get benefits, you will need to give documents that show things like how much money you get, costs to live, and where you live.<br/><br/>Adding them now can help the process go faster. <br/> <br/><b>If you don't know what to add, skip it and you can talk with a county worker later.</b>
 doc-upload-intro.continue=Yes, upload now
 doc-upload-intro.skip=I'll do this later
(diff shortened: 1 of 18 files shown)


src/main/resources/messages.properties
```diff
@@ -1018,7 +1018,7 @@ ethnicity-selection.option1=Hispanic or Latino
 ethnicity-selection.option2=Not Hispanic or Latino
 
 race-selection.title=Race Selection
-race-selection.header=What races do you ({0}) identify with?
+race-selection.header=What race`s do you ({0}) identify with?
```
bseeger, line 1021: Changing this back. Not sure what happened here, but it's def. not right.

vrajmohan: Hmm...if we expect only one to be chosen, it should be "race", else it should be "races". "race's" is wrong.

src/test/java/org/mdbenefits/app/utils/TestUserFileUtilities.java
```diff
@@ -0,0 +1,18 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+public class TestUserFileUtilities {
+    @CsvSource(delimiter=';', value = {
+        // passed in value; expected value
+        ".jpg, .jpeg, .csv,.pdf;.jpg, .jpeg, .csv, and .pdf",
```
vrajmohan, line 11: Feel silly mentioning it, but as we are doing joins and Oxford commas, I would add cases for 1 file type and 2 file types.

vrajmohan (approved):
Minor comment

src/test/java/org/mdbenefits/app/utils/TestUserFileUtilities.java
bseeger: Oh, that's a very good point.

bseeger: I added code to address this.  I think it was good to tighten this up some.

vrajmohan (approved):
Looks great

PR #37, opened 2024-02-06
Implements Need and Program Pages

analoo:
Implements Need and Program Pages:

To test:
1. Select A Need (if you select either of the top two options, the next window will show both options with the need you selected prechecked).
2. Go Back - Select all needs (except not sure) and confirm that in the next page, 4 options appear that are all checked.
3. Go back. Select "Not Sure". Notice it act's as a "None of the Above". Go to the next page and note that all options are listed but none are checked.

Play around with moving forward to the signpost, and then back: Does the behavior make sense to you?

Once you feel good about the need, check that the pdf form is properly filled out:
If you select SNAP, the snap option on the first page will be checked.
If you select something aside from SNAP, Cash Assistance will be checked.


src/main/java/org/mdbenefits/app/data/enums/HelpNeededType.java
```diff
@@ -3,6 +3,7 @@
 import static org.mdbenefits.app.data.enums.ProgramType.SNAP;
 import static org.mdbenefits.app.data.enums.ProgramType.TCA;
 import static org.mdbenefits.app.data.enums.ProgramType.TDAP;
+import static org.mdbenefits.app.data.enums.ProgramType.RCA;
 import static org.mdbenefits.app.data.enums.ProgramType.OTHER;
```
analoo, line 7: To do: make "other option" a "none of the above"

src/main/java/org/mdbenefits/app/submission/actions/PreCheckPrograms.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ *
+ */
+@Component
+@Slf4j
+public class PreCheckPrograms implements Action {
+
+    @Override
+    public void run(Submission submission) {
+        ArrayList<String> helpNeeded = (ArrayList<String>) submission.getInputData().get("helpNeeded[]");
```
analoo, line 18: to do: when the user selects "Other, default to include all of the need types

src/main/resources/flows-config.yaml
```diff
@@ -15,6 +15,7 @@ flow:
     nextScreens:
       - name: redirectToMDThink
   selectHelpNeeded:
+    beforeSaveAction: PreCheckPrograms
```
analoo, line 18: @bseeger - any strong sentiment between adding this as a beforeSave action or a before display action on the next page?

src/main/resources/templates/mdBenefitsFlow/choosePrograms.html
```diff
@@ -14,18 +14,20 @@
             th:replace="~{fragments/form :: form(action=${formAction}, content=~{::formContent})}">
           <th:block th:ref="formContent">
             <div class="form-card__content">
+              <h2 th:text="#{choose-programs.check-all-that-apply}"></h2>
               <th:block th:replace="~{fragments/inputs/checkboxFieldset ::
                   checkboxFieldset(inputName='programs',
-                  label=#{choose-programs.check-all-that-apply},
+                  ariaLabel=#{choose-programs.header},
```
analoo, line 20: @bseeger - do we have a best practice around what to include in the ariaLabel?

src/main/java/org/mdbenefits/app/data/enums/HelpNeededType.java
analoo: Since "other" will act as none of the above, it might make sense to remove from the HelpNeededType.

I can add a separate checkBoxInSet that is noneOfTheAbove, and therefore saves as an empty list. 

On PreCheckPrograms, we can skip the logic if HelpNeeded is empty.

Then in the chosePrograms template, I can add a different logic that uses all of the HelpNeededType when program is empty.

OR

There can be another screen chooseProgramsOther that always displays everything on the HelpNeededType.

src/main/java/org/mdbenefits/app/submission/actions/PreCheckPrograms.java
```diff
@@ -0,0 +1,28 @@
...
+        ArrayList<String> programsToAdd = new ArrayList<>();
+
+        helpNeeded.forEach((String need) -> {
```
analoo, line 21: To do: Confirm that if help needed is an empty list, it won't create any errors.

aka, it will just save programs = []

src/main/resources/flows-config.yaml
bseeger: I think it was already checking them based on the value of the help needed...?

src/main/java/org/mdbenefits/app/submission/actions/PreCheckPrograms.java
bseeger: Overall, I think it might be better to do the preselect in the page flow rather than here, in an action.   You can key off the "helpNeeded[]" values to set some to preselect.

src/main/resources/templates/mdBenefitsFlow/choosePrograms.html
bseeger: no, I was wondering that as well. Might be worth checking in with Devon about this.

src/test/java/org/mdbenefits/app/pdf/PdfServiceTest.java
```diff
@@ -37,6 +37,7 @@ void generate() throws IOException {
                 .with("homeAddressState", "CA")
                 .with("homeAddressZipCode", "94103")
                 .with("phoneNumber", "(510) 555-1212")
+                .with("programs[]", "['SNAP', 'RCA']")
```
analoo, line 40: Looks like I need a "List.of()" here

src/main/java/org/mdbenefits/app/data/enums/HelpNeededType.java
```diff
@@ -3,24 +3,45 @@
...
+            MAP_BY_PROGRAM_TYPE.put(type.programType.toString(), type.relevantPrograms);
+        }
+    }
 
-  private final String labelSrc;
-  private final ProgramType programType;
+    HelpNeededType(String labelSrc, ProgramType programType, ArrayList<ProgramType> relevantPrograms) {
+        this.labelSrc = labelSrc;
+        this.programType = programType;
+        this.relevantPrograms = new ArrayList<>(relevantPrograms);
```
analoo, line 40: remove new Arraylist since it is being created above.

src/main/resources/templates/mdBenefitsFlow/selectHelpNeeded.html
```diff
@@ -20,14 +20,21 @@
                   label=#{general.check-all-that-apply},
                   content=~{::content})}">
                 <th:block th:ref="content">
-                  <th:block th:each="helpNeededType : ${T(org.mdbenefits.app.data.enums.HelpNeededType).values()}">
+                  <th:block
+                      th:each="helpNeededType : ${T(org.mdbenefits.app.data.enums.HelpNeededType).values()}">
                     <th:block
                         th:replace="~{fragments/inputs/checkboxInSet :: checkboxInSet(
                             inputName='helpNeeded',
                             value=${helpNeededType.getProgramType()},
```
analoo, line 28: helpNeededType.FOOD.name()

src/main/java/org/mdbenefits/app/data/enums/HelpNeededType.java
```diff
@@ -3,24 +3,45 @@
...
+    }
 
-  HelpNeededType(String labelSrc, ProgramType programType) {
-    this.labelSrc = labelSrc;
-    this.programType = programType;
-  }
+    public static ArrayList getRelevantProgramsFromValue(String programType) {
+        ArrayList<ProgramType> relevantPrograms = MAP_BY_PROGRAM_TYPE.get(programType);
+        return relevantPrograms;
+    }
```
analoo, line 46: We could have something that you can provide the name and it returns the programValues

Map By ENUM Name

src/main/java/org/mdbenefits/app/submission/actions/PreCheckPrograms.java
```diff
@@ -0,0 +1,39 @@
...
+    @Override
+    public void run(Submission submission) {
+        ArrayList<String> helpNeeded = (ArrayList<String>) submission.getInputData().get("helpNeeded[]");
+        ArrayList<String> programsToAdd = new ArrayList<>();
+
+        if (helpNeeded.contains("None")) {
+            submission.getInputData().remove("programs[]");
+        } else {
+            helpNeeded.forEach((String need) -> {
+                ArrayList<ProgramType> relevantPrograms = (ArrayList<ProgramType>) HelpNeededType.getRelevantProgramsFromValue(need);
```
analoo, line 27: potentialPrograms could be a utility method that returns a flattened list of the HelpNeededType programTypes. We can use this flattened list to iterate on the choose Programs page.

This action stays as a way to set the checks for certain programs.

src/main/java/org/mdbenefits/app/data/enums/HelpNeededType.java
```diff
@@ -3,24 +3,55 @@
...
 
 @Getter
 public enum HelpNeededType {
 
-  FOOD( "help-needed.food", SNAP),
-  CHILDREN("help-needed.children", TCA),
-  //UTILITIES("help-needed.utilities", OHEP),
-  DISABILITY("help-needed.disability", TDAP),
-  NOT_SURE("help-needed.not-sure", OTHER);
+    FOOD("help-needed.food", SNAP, new ArrayList<>(List.of(SNAP, TCA))),
```
analoo, line 18: Kept Program Type as the key that determines what we would want to have pre-checked. relevant programs determines the potential options that will appear to the final user.

src/main/java/org/mdbenefits/app/submission/actions/PreCheckPrograms.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.ArrayList;
+
+import lombok.extern.slf4j.Slf4j;
+import org.mdbenefits.app.data.enums.HelpNeededType;
+import org.mdbenefits.app.utils.ApplicationUtilities;
+import org.springframework.stereotype.Component;
+
+/**
+ *
+ */
```
bseeger, line 15: Can  you fill in this comment?  Thanks!

src/main/java/org/mdbenefits/app/preparers/ApplicantNeedsPreparer.java
```diff
@@ -0,0 +1,38 @@
...
+@Component
+public class ApplicantNeedsPreparer implements SubmissionFieldPreparer {
+
+        @Override
+        public Map<String, SubmissionField> prepareSubmissionFields(Submission submission, PdfMap pdfMap) {
+                Map<String, SubmissionField> results = new HashMap<>();
+
+                Map<String, Object> inputData = submission.getInputData();
+
+                ArrayList<String> programs = (ArrayList) inputData.get("programs[]");
```
bseeger, line 23: For the declaration part, you could just use List, as it's the parent type for ArrayList(). 
Not really necessary, but kinda handy - like the way on line 19 where you just declare "Map<String, SubmissionField>", but it's really a HashMap. 

`List<String> programs = (List) inputData.get("programs[]");` should work.  We don't actually know that it's an ArrayList, and really we don't need to care about that - we only care that it behaves like a list.

src/main/java/org/mdbenefits/app/submission/actions/PreCheckPrograms.java
```diff
@@ -0,0 +1,35 @@
...
+@Component
+@Slf4j
+public class PreCheckPrograms implements Action {
+
+        @Override
+        public void run(Submission submission) {
+                ArrayList<String> helpNeeded = (ArrayList<String>) submission.getInputData().get("helpNeeded[]");
```
bseeger, line 22: Here is where I'd just use List again: 

`List<String> helpNeeded = (List) submission.getInputData().get("helpNeeded[]");`

src/main/java/org/mdbenefits/app/submission/actions/PreCheckPrograms.java
```diff
@@ -0,0 +1,35 @@
...
+                ArrayList<String> programsToAdd = new ArrayList<>();
```
bseeger, line 23: Here, just use: 

`List<String> programsToAdd = new ArrayList<>();`

src/main/java/org/mdbenefits/app/utils/ApplicationUtilities.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+import org.bouncycastle.util.Strings;
+import org.mdbenefits.app.data.enums.HelpNeededType;
+import org.springframework.shell.Command.Help;
+
+public class ApplicationUtilities {
+
+        public static Set<String> flattenHouseholdNeedsPrograms(ArrayList<String> needTypes) {
+                ArrayList<ArrayList<String>> relevantPrograms = new ArrayList<>();
```
bseeger, line 19: I think you could just do `List<List<String>>` here for the left side of the equals.

src/main/resources/templates/mdBenefitsFlow/choosePrograms.html
```diff
@@ -14,22 +14,29 @@
...
+                        th:each="programType : ${T(org.mdbenefits.app.data.enums.ProgramType).values()}">
                       <th:block
-                          th:replace="~{fragments/inputs/checkboxInSet:: checkboxInSet(
+                          th:if="${relevantPrograms.contains(programType)}">
+                        <h2
+                            th:with="shortDesc=${programType.getShortDescSrc().isBlank()} ? '' : #{${programType.getShortDescSrc()}}"
+                            th:utext="${shortDesc}">
+                        </h2>
+                        <th:block class="font-weight--normal"
+                                  th:replace="~{fragments/inputs/checkboxInSet:: checkboxInSet(
```
bseeger, line 34: I don't think that putting `class="font-weight--normal"` here will do anything, as the `th:block` is a construct that's not actually put into the HTML.  The class needs to go on whatever the "checkboxInSet" produces.   You'd need to edit this code, though it already may be done (but not 100% sure it's right). https://github.com/codeforamerica/mdpilot/blob/main/src/main/resources/templates/fragments/inputs/checkboxInSet.html

bseeger: Try taking out the class stuff on this th:block and see how it renders.

src/test/java/org/mdbenefits/app/journeys/MdBenefitsFlowJourneyTest.java
```diff
@@ -15,941 +15,1006 @@
...
+                testPage.navigateToFlowScreen("mdBenefitsFlow/selectApplication");
+                testPage.clickElementById("none__checkbox");
+                testPage.clickContinue();
+                assertThat(testPage.getTitle()).isEqualTo(message("help-needed.title"));
+        }
+
+        @Test
+        void redirectToMyMDTHINKOnUnsupportedCounty() {
+                testPage.navigateToFlowScreen("mdBenefitsFlow/county");
+                // should redirect to MyMDTHINK
```
bseeger, line 42: I'm guessing you meant to only have 4 space indents here?  Can you revert the indent change?  Thanks!

bseeger:
@analoo, this looks really great and runs well!  I just have some minor comments.

src/test/java/org/mdbenefits/app/journeys/MdBenefitsFlowJourneyTest.java
bseeger: I haven't fully read this file yet as the diff is noisy.  I'll read it on Monday.

src/main/java/org/mdbenefits/app/preparers/ApplicantNeedsPreparer.java
analoo: I did not know that List is the parent type of arrayList. Thank you for sharing! I'll make these changes

src/main/resources/templates/mdBenefitsFlow/choosePrograms.html
analoo: You are correct. I removed the font weight class and it didn't affect anything. Removing it.

src/main/java/org/mdbenefits/app/data/enums/HelpNeededType.java
```diff
@@ -3,24 +3,55 @@
...
+    NOT_SURE("help-needed.not-sure", null, new ArrayList<>(List.of(RCA, TDAP, SNAP, TCA)));
+
+    private final String labelSrc;
+    private final ProgramType programType;
+    private final ArrayList<ProgramType> relevantPrograms;
+
+    static private final Map<String, ArrayList<ProgramType>> MAP_BY_ENUM_NAME = new HashMap<>();
+    static private final Map<String, ProgramType> MAP_PROGRAM_BY_ENUM_NAME = new HashMap<>();
+
+    static {
```
bseeger, line 31: I think you can combine these static blocks into one.

src/main/java/org/mdbenefits/app/data/enums/HelpNeededType.java
```diff
@@ -3,24 +3,55 @@
...
+
+    HelpNeededType(String labelSrc, ProgramType programType, ArrayList<ProgramType> relevantPrograms) {
+        this.labelSrc = labelSrc;
+        this.programType = programType;
+        this.relevantPrograms = relevantPrograms;
+    }
 
-  private final String labelSrc;
-  private final ProgramType programType;
+    public static ArrayList getRelevantProgramsFromName(String name) {
```
bseeger, line 48: Sorry I didn't note this before. This return value should be a List as well.

src/main/java/org/mdbenefits/app/utils/ApplicationUtilities.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import org.bouncycastle.util.Strings;
+import org.mdbenefits.app.data.enums.HelpNeededType;
+import org.springframework.shell.Command.Help;
+
+public class ApplicationUtilities {
+        public static Set<String> flattenHouseholdNeedsPrograms(ArrayList<String> needTypes) {
+                List<List<String>> relevantPrograms = new ArrayList<>();
```
bseeger, line 18: The indents are off in this file -- 8 instead of 4.

src/test/java/org/mdbenefits/app/journeys/MdBenefitsFlowJourneyTest.java
bseeger: Thank you!

src/test/java/org/mdbenefits/app/journeys/MdBenefitsFlowJourneyTest.java
```diff
@@ -52,33 +52,101 @@ void redirectToMyMDTHINKOnUnsupportedCounty() {
...
-    void selectHelpNeededAndChooseProgramFlow() {
+    void selectNotSureHelpNeededAndChooseProgramFlow() {
         // select help needed
         testPage.navigateToFlowScreen("mdBenefitsFlow/selectHelpNeeded");
         testPage.clickContinue();
         assertThat(testPage.getTitle()).isEqualTo("Select help");
-        assertThat(testPage.hasErrorText(message("error.missing-general")));
+        assert(testPage.hasErrorText(message("error.missing-general")));
+
+        testPage.clickElementById("helpNeeded-CHILDREN");
```
bseeger, line 62: I think you can take this line out -- or comment that it does nothing, as the checking `none__checkbox` will cause it to get unchecked.   And maybe that's part of the test as well?

analoo: - I added a comment since I believe that this is actually an important part of the test since none__checkbox unsets the other checkboxes and behaves differently in the programs pages.

src/main/resources/messages.properties
```diff
@@ -246,6 +246,7 @@ help-needed.food=Help with food
 help-needed.children=Help with money for children
```
bseeger, line 246: It won't let me mark the right spot, so I'll put this here.  If this PR includes the help needed page, then the message for `help-needed.subheader` needs to be updated as it no longer matches the design.

src/main/resources/messages.properties
```diff
@@ -266,6 +267,9 @@ choose-programs.ohep.desc=<br>Helps with heating, electric and gas bills. Maryla
 choose-programs.tdap=Temporary Disability Assistance Program (TDAP)
 choose-programs.tdap.short-desc=Get help for a disability
 choose-programs.tdap.desc=<br>Gives cash to people who are low-income and living with short-term disability or who are waiting for federal disability help. People with children who are dependents cannot get this benefit.
+choose-programs.rca=Refugee Transitional Cash Assistance (RTCA or RCA)
```
bseeger, line 270: Note that it looks like the copy for SNAP and TCA has changed in the designs. 

The TDAP needs a line break in on  spot to match design.

bseeger (approved):
This works as described. There are a few copy changes - feel free to merge once you make those. 

Nice work - the change to the pdf is nice!

src/main/resources/templates/mdBenefitsFlow/choosePrograms.html
spokenbird: It should be the ID of the element you want to label the input. It's meant for screens where there is only one input, and the header is technically the label for that input. In most cases it will be 'header' since that's the ID assigned to headers in the FFB.

PR #40, opened 2024-02-08
Send email to applicant on signing

vrajmohan:


vrajmohan:
I will update the deploy pipeline to update Mailgun config vars in a separate commit.

src/main/java/org/mdbenefits/app/submission/actions/HandleApplicationSigned.java
```diff
@@ -1,16 +1,55 @@
...
+    }
+
     @Override
     public void run(Submission submission) {
+        sendEmailToApplicant(submission);
+    }
+
+    private void sendEmailToApplicant(Submission submission) {
+        String recipientEmail = (String) submission.getInputData().get("emailAddress");
+        if (recipientEmail == null || recipientEmail.isBlank()) {
```
analoo, line 36: @vrajmohan - How would you feel about adding a check here to see what environment we are in. 

If, in staging or dev, maybe we can skip email sends, or override to send to a test mailing list instead?

We are testing a lot and I'm concerned that we might accidentally send emails out every once in a while.

src/main/resources/application.yaml
```diff
@@ -38,6 +38,11 @@ form-flow:
       smarty_auth_id: ${SMARTY_AUTH_ID}
       smarty_auth_token: ${SMARTY_AUTH_TOKEN}
       smarty_url: "https://us-street.api.smartystreets.com/street-address"
+  email-client:
```
analoo, line 41: @vrajmohan - does this configuration only apply to the production app? If so, please disregard this comment: https://github.com/codeforamerica/mdpilot/pull/40/files#r1483722700

src/main/java/org/mdbenefits/app/submission/actions/HandleApplicationSigned.java
vrajmohan: Good point, see comment below.

src/main/resources/application.yaml
vrajmohan: The configuration applies to all environments but relies on the `MAILGUN_KEY` env var being set. The email will not be sent if `MAILGUN_KEY` is incorrect or not set.

analoo (approved):

PR #42, opened 2024-02-10
Update display icon to checkmark in pdf checkboxes

analoo:
https://www.pivotaltracker.com/story/show/187021846

Changes to Highlight:
1) Multiple Boolean checkboxes that are related are combined into a radio button group. The value of the radio button is true for yes and false for no.
2) Multiple Checkboxes (more than 2 or non boolean) that are related are combined into a radio button group. The value of the radio button is the name of the field. 
3) All radio button values and checkboxes will display as red checkmarks in size 8
4) Where name of a field made no sense, it was updated. Where name of a field was a sentence, it was updated to a single string.


bseeger:
Hi Ana, can you put a generic description in the title (`update PDF template`) and then list out details in the description section above?  Thanks!!

PR #46, opened 2024-02-13
Add address validation flow

spokenbird:
This PR utilizes JavaScript on the mailing address screen within the thymeleaf template to grab input values from the home address screen if the user selects that their mailing address is the same as their home address. It then inserts those values in the mailing address inputs so that Smarty has the address fields to validate within the form data at the time the POST occurs.


src/main/resources/templates/fragments/inputs/singleCheckbox.html
```diff
@@ -0,0 +1,43 @@
...
+        (#arrays.length(errorMessages.get(inputName)) > 0) }"
+    th:assert="
+      ${!#strings.isEmpty(inputName)},
+      ${!#strings.isEmpty(value)},
+      ${!#strings.isEmpty(label)}">
+  <div th:class="'form-group' + ${(hasError ? ' form-group--error' : '')}">
+    <label th:for="${inputName} + '-' + ${value}"
+           th:id="${inputName} + '-' + ${value} + '-label'"
+           class="checkbox display-flex">
+<!--      <input type="hidden" th:id="${inputName} + 'Hidden'" th:name="${name}" value="">-->
```
bseeger, line 19: do you need this anymore?

src/main/resources/templates/mdBenefitsFlow/mailingAddress.html
```diff
@@ -41,5 +41,38 @@
...
+    if (checkbox && checkbox.checked) {
+      // Copy values from home address to mailing address
+      document.getElementById("mailingAddressStreetAddress1").value = inputData['homeAddressStreetAddress1'] || '';
+      document.getElementById("mailingAddressStreetAddress2").value = inputData['homeAddressStreetAddress2'] || '';
+      document.getElementById("mailingAddressCity").value = inputData['homeAddressCity'] || '';
+      document.getElementById("mailingAddressState").value = inputData['homeAddressState'] || '';
+      document.getElementById("mailingAddressZipCode").value = inputData['homeAddressZipCode'] || '';
+    } else {
+      if (!onPageLoad) {
+        // Clear mailing address fields if checkbox is unchecked
```
bseeger, line 59: I think you need to consider "onPageLoad" for the checkbox as well.

I am seeing my corrected mailing address  (was same as home address, so the checkbox is clicked) getting overwritten again by the home address, when I go back to the page. 

Scenario: I enter a home address and submit it; then on the mailing page I click that my mailing address is the same as my home address. On the next page I choose the corrected version and click continue.  Then I go back to the mailing address page and see the old (uncorrected) home address.  

The only time the home address should override the mailing is if I actively click on the box on that - not if it's set when the page loads.

bseeger (commented):
This looks good overall, but there is one spot I called out that might need a change.

PR #48, opened 2024-02-14
Remove unnecessary pages

bseeger:
Ticket: https://www.pivotaltracker.com/story/show/187035371

Removes the following pages: 
 * Foster child y/n
 * Anyone in foster care until they turned 18?
 * Currently experiencing homelessness?
 * Rent a room?
 * Does anyone need to get away from an abusive situation
 
Updates relevant journey tests.


src/main/resources/messages.properties
```diff
@@ -1047,10 +1031,6 @@ doc-upload-signpost.well-ask-about=We'll tell you
 doc-upload-signpost.how-to-add-docs=How to add docs
 doc-upload-signpost.which-docs=Which docs you can add
 doc-upload-signpost.how-to-submit=How to submit now or later
-# doc upload recommendations
```
spokenbird, line 1050: Assuming these are being removed intentionally but noting that it doesn't seem like these message properties were part of any screens that were removed?

spokenbird (approved):
lgtm

PR #49, opened 2024-02-15
Prepares flow for upcoming applicant pages

bseeger:
This creates the pages that will be in upcoming tickets, as well as updates the page flow. Tests have been updated as well to prepare for upcoming changes.

Addresses:  https://www.pivotaltracker.com/story/show/187042163 & https://www.pivotaltracker.com/story/show/187054131 (super small page)

Adds the first and last page as a convenience and so testing flows well.


src/main/resources/flows-config.yaml
```diff
@@ -97,20 +95,63 @@ flow:
...
-      - name: householdInfo
-  householdInfo:
-    crossFieldValidationAction: ValidateHouseholdBirthdate
-    subflow: household
+      - name: applicantApplying
+  applicantApplying:
+    nextScreens:
+      - name: applicantSex
+        condition: IsApplicantApplyingForSelf
+      - name: householdHasHousehold
```
vrajmohan, line 109: Minor: rename to `householdHasAdditionalMembers`?

vrajmohan (approved):
Made a minor comment. Only did a cursory review as the main goal is to to avoid stepping on each other. Real issues can be resolved as we implement the dependent stories.

PR #53, opened 2024-02-16
Updating logging behavior

spokenbird:


src/main/java/org/mdbenefits/app/submission/actions/HandleApplicationSigned.java
```diff
@@ -44,7 +44,7 @@ private void sendEmailToApplicant(Submission submission) {
                 messageBody
         );
         if (mailgunResponse != null) {
-            log.info("Sent email. Mailgun response: {}", mailgunResponse.getMessage());
+            log.info("Sent email for application with ID: {}. Mailgun response: {}", submission.getId(), mailgunResponse.getMessage());
```
vrajmohan, line 47: I didn't bother adding this as it is in the MDC.

vrajmohan (approved):
Minor comment that need not be addressed

src/main/java/org/mdbenefits/app/submission/actions/HandleApplicationSigned.java
spokenbird: Good point!

PR #59, opened 2024-02-20
Remove uneccesary pages follow up

analoo:
Remove OutOfState Household benefits page: https://www.pivotaltracker.com/story/show/187035371/comments/240250657

Update flow so that after review information, we move to multiplePersonHousehold https://www.pivotaltracker.com/story/show/187054131/comments/240268405


src/main/java/org/mdbenefits/app/inputs/MdBenefitsFlow.java
```diff
@@ -206,22 +202,6 @@ public class MdBenefitsFlow extends FlowInputs {
 
     private String veterans;
 
-    private String fosterInd;
```
analoo, line 209: I realized these were not removed from Bethany's original PR

spokenbird (approved):
looks good!

PR #60, opened 2024-02-20
Implement Applicant Applying screen

spokenbird:


analoo (approved):
Code LGTM. I checked out the PR and confirmed that the designs match the implementation (with the exception of the required asterisk which is not currently available).

PR #62, opened 2024-02-21
Implement individual school screen

analoo:
Link to Pivotal: https://www.pivotaltracker.com/story/show/187048418


src/test/java/org/mdbenefits/app/journeys/MdBenefitsFlowJourneyTest.java
```diff
@@ -685,6 +677,15 @@ void completeFlow() {
         testPage.clickElementById("preparesFood-you");
         testPage.clickContinue();
 */
+        testPage.navigateToFlowScreen("mdBenefitsFlow/applicantApplying");
```
spokenbird, line 680: The code looks good, approved, but I do wonder if we should move this logic into a helper method for the test? Something like completeApplicantInformationSection() that does all this logic around different potential paths / conditional logic tree navigation?

spokenbird (approved):
Looks good!

src/test/java/org/mdbenefits/app/journeys/MdBenefitsFlowJourneyTest.java
analoo: that's a good point. For now, I added a fork in the testing so that we could explore an individual flow and a family flow.  Let's discuss tomorrow in standup and I can shift in a follow up PR.

PR #70, opened 2024-02-26
Upgrade springframework gradle plugin

bseeger:
After this [PR](https://github.com/codeforamerica/mdpilot/pull/69) is in, this one should work. 

This upgrades the `org.springframework.boot` gradle plugin to 3.2.3. 

Works with gradle version 8.6


bseeger:
I'll pick this up again in the morning. It worked locally with the changes, but looks like it needs a little more work. :)

PR #76, opened 2024-02-28
Generate confirmation number

vrajmohan:


analoo (approved):
LGTM.

src/main/java/org/mdbenefits/app/submission/actions/HandleApplicationSigned.java
```diff
@@ -19,15 +20,21 @@ public class HandleApplicationSigned implements Action {
...
         this.submissionRepositoryService = submissionRepositoryService;
+        this.jdbcTemplate = jdbcTemplate;
     }
 
     @Override
     public void run(Submission submission) {
+        String confirmationNumber = jdbcTemplate.queryForObject("select 'M-' || nextval('confirmation_sequence')", String.class);
+        submission.getInputData().put("confirmationNumber", confirmationNumber);
+        log.info("Using confirmationNumber: {}", confirmationNumber);
         sendEmailToApplicant(submission);
```
bseeger, line 38: This does not work as expected. We need to explicitly save the submission again here, else the confirmation number will be gone and not actually stored in the data.

It is not saved automatically as the submission object is not considered "managed" anymore (I think). 

What I did was run this code w/o an email recipient.  The confirmation number was lost. 

When I ran with a recipient address, the confirmation number is save correctly, but only because the section that sends the email will call save explicitly.

bseeger (changes requested):
Looks good and like what we talked about.  There is one thing that needs to be changed before this will function as expected.

src/main/java/org/mdbenefits/app/submission/actions/HandleApplicationSigned.java
```diff
@@ -19,18 +20,29 @@ public class HandleApplicationSigned implements Action {
...
     @Override
     public void run(Submission submission) {
+        assignConfirmationNumber(submission);
         sendEmailToApplicant(submission);
     }
 
+    private void assignConfirmationNumber(Submission submission) {
+        String confirmationNumber = jdbcTemplate.queryForObject("select 'M-' || nextval('confirmation_sequence')", String.class);
+        submission.getInputData().put("confirmationNumber", confirmationNumber);
+        log.info("Using confirmationNumber: {}", confirmationNumber);
```
bseeger, line 42: I like what you did with this. Can you add the submission ID to this log message? I could see that being useful during debug (rather than having to look it up separately).

bseeger (approved):
One minor request. Looks good overall!

src/main/java/org/mdbenefits/app/submission/actions/HandleApplicationSigned.java
vrajmohan: I didn't add it as it is in the MDC.

bseeger: Oh, I forgot about that.  Never mind then.  :)

PR #79, opened 2024-02-28
Make all race fields multi-line to handle possible length of data.

bseeger:
This makes all the fields in the Race column multi-line, in case someone picks many of the options. 

I don't know how often all 5 will be chosen, so maybe this isn't big enough:

Screenshot 2024-02-28 at 9 50 39 AM


bseeger:
Redoing to make the field wider

PR #81, opened 2024-02-28
Adds Applicant Race and Ethnicity Page

bseeger:
Addresses: https://www.pivotaltracker.com/story/show/187048663


src/main/resources/pdf-map.yaml
```diff
@@ -45,96 +47,89 @@ inputFields:
...
-    Money Market Account: Money Market Account
-    Mutual funds: Mutual funds
-    Savings bond: Savings bond
-    Stocks: Stocks
-
-  needsNewEbtCard: needsNewEbtCard
-
-  votingRegistrationRequested: votingRegistrationRequested
-  votingRegistrationHelpRequested: votingRegistrationHelpRequested
+  #  citizenshipStatus: applicant_is_citizen
```
bseeger, line 52: because these were under the `dbFields` category, there were a bunch of error messages. I just commented these out to quiet that down, until we decide if we need them anymore.

src/test/java/org/mdbenefits/app/pdf/PdfServiceTest.java
```diff
@@ -25,38 +27,46 @@ class PDFServiceTest {
...
+        assertThat(page2).contains("2,5");
     }
 
-    private static String getText(File file) throws IOException {
+    private static String getPageText(File file, int page) throws IOException {
         try (PdfReader reader = new PdfReader(file.getPath())) {
             PdfTextExtractor pdfTextExtractor = new PdfTextExtractor(reader);
-            return pdfTextExtractor.getTextFromPage(1);
+            return pdfTextExtractor.getTextFromPage(page);
         }
```
bseeger, line 70: Not crazy about the time it takes to run the extractor, just to extract a page at a time, but I can't think of a better way to do this right now and it's probably not a big hit anyways.

src/main/resources/pdf-map.yaml
analoo: This is great! @spokenbird and I commented out outdated fields as well! So I think we are on the same page here.

analoo (approved):
LGTM.

Heads up there may be some merge conflicts with the household work that Alex and I put together but they may be pretty minor.

bseeger:
Working on resolving conflicts. I'll have this ready in the morning.

PR #82, opened 2024-02-28
Add Mailgun configuration for prod

bseeger:
Adds configuration for the production server to send email via mailgun.


bseeger:
Paired with @vrajmohan to do this work.

PR #84, opened 2024-02-28
Update Applicant Disability Page with writing to PDF

analoo:


src/main/java/org/mdbenefits/app/preparers/ApplicantDetailsPreparer.java
```diff
@@ -48,6 +48,10 @@ public Map<String, SubmissionField> prepareSubmissionFields(Submission submissio
             if (inputData.get("isApplicantPregnant").toString().equalsIgnoreCase("true")) {
                 results.put("applicantIsPregnantName", new SingleField("applicantIsPregnantName", (String) fullName, null));
             }
+
+            if (inputData.get("applicantHasDisability").toString().equalsIgnoreCase("Yes")) {
+                results.put("applicantHasDisabilityName", new SingleField("applicantHasDisabilityName", (String) fullName, null));
+            }
```
bseeger, line 54: for most of these if we use `inputData.getOrDefault` with a reasonable default value, then these will not throw an error if the field is unset. I know our happy path through the app has these set, but if we, as devs, want to jump ahead and created a PDF (like jumping to the `signature` page), this logic will throw an error and the rest of this preparer's logic won't run.

PR #88, opened 2024-02-29
Update copy for job triage screen

analoo:
https://www.pivotaltracker.com/story/show/187055710

Instead of updating, I had to create because this was one of the pages removed post UX/Design/Product chat

### DIFF
--- a/src/main/resources/templates/mdBenefitsFlow/householdIncome.html
+++ b/src/main/resources/templates/mdBenefitsFlow/householdIncome.html
@@ -23,6 +23,7 @@
                 <li th:text="#{household-income.li3}"></li>
                 <li th:text="#{household-income.li4}"></li>
                 <li th:text="#{household-income.li5}"></li>
+                <li th:text="#{household-income.li6}"></li>
               </ul>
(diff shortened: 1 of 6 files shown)


src/main/resources/templates/mdBenefitsFlow/householdIncome.html
```diff
@@ -0,0 +1,41 @@
...
+          <th:block th:ref="formContent">
+            <div class="form-card__content">
+              <p th:text="#{household-income.p1}"></p>
+              <p th:text="#{household-income.p2}"></p>
+              <ul class="list--bulleted">
+                <li th:text="#{household-income.li1}"></li>
+                <li th:text="#{household-income.li2}"></li>
+                <li th:text="#{household-income.li3}"></li>
+                <li th:text="#{household-income.li4}"></li>
+                <li th:text="#{household-income.li5}"></li>
```
bseeger, line 25: Just noting that the li6 is not showing up here.  The line about the barber, construction, etc... 

Screenshot 2024-03-01 at 5 47 00 PM

PR #99, opened 2024-03-04
Change copy on address-related screens

vrajmohan:


src/main/resources/templates/mdBenefitsFlow/mailingAddress.html
```diff
@@ -24,8 +24,7 @@
                 <th:block th:replace="~{fragments/inputs/address ::
                                   address(
                                     validate=true,
-                                    inputName='mailingAddress',
-                                    streetAddress2HelpText=#{general.optional}
+                                    inputName='mailingAddress'
```
bseeger, line 27: Is there a compelling technical reason to have two separate pages for the scenarios?  From what I can tell, the only difference between this and the `mailingAddressNoAddresse` page is some JS and the logic to add the checkbox to use homeAddress. 

I realize this was in before you worked on this ticket, but do you know of a good reason to not just have these two be one page where the checkbox is conditionally shown?

src/main/resources/templates/mdBenefitsFlow/homeAddress.html
```diff
@@ -10,7 +10,8 @@
       <main id="content" role="main" class="form-card spacing-above-35">
         <th:block th:replace="~{fragments/icons :: county}"></th:block>
```
bseeger, line 11: This icon should be `mail`.

src/main/resources/templates/mdBenefitsFlow/mailingAddress.html
vrajmohan: I was thinking the very same thing and decided to stick with the existing pattern so as to minimize change.

src/main/resources/templates/mdBenefitsFlow/homeAddress.html
vrajmohan: Good catch. I hadn't spotted the difference at all.

vrajmohan:
While we are at it, I was unhappy with the field name `sameAsHomeAddress` as it is meaningless without context. I wanted to change it to `mailingAddressSameAsHomeAddress` but held back as that would be changes in many places.

src/main/resources/templates/mdBenefitsFlow/mailingAddressNoAddress.html
```diff
@@ -19,8 +19,7 @@
                 <th:block th:replace="~{fragments/inputs/address ::
                                   address(
                                     validate=true,
-                                    inputName='mailingAddress',
-                                    streetAddress2HelpText=#{general.optional}
+                                    inputName='mailingAddress'
```
bseeger, line 22: I thought we talked about keeping this (at least as the smaller group)?  Or was there further discussion?

src/main/resources/templates/mdBenefitsFlow/mailingAddress.html
```diff
@@ -24,8 +24,7 @@
...
                                   )}"/>
```
bseeger, line 28: I thought we talked about keeping the optional (at least as a smaller group).   Was there more conversation on this?

src/main/resources/templates/mdBenefitsFlow/mailingAddressNoAddress.html
vrajmohan: There was no further discussion and I was going to bring it up today.

bseeger:
One more note -- if you don't have a permanent address, but add a mailing address it does not show up on the Review Page.  Not sure if this PR breaks that, but just FYI.  I think that whole page needs to get re-done at some point anyways, but if it points to a problem with where the address is stored, then we should fix it.

src/main/resources/templates/mdBenefitsFlow/mailingAddress.html
bseeger: Okay, that works for me.  For now we can leave it.

bseeger (approved):
This looks good to me once the icon is fixed. 

We could wait to merge until we answer the question of the address second line.

src/main/resources/pdf-map.yaml
```diff
@@ -68,130 +68,3 @@ subflowInfo:
...
-#    inputFields:
-#      studentName: studentName
-#      schoolName: schoolName
-#      attendance: attendance
-#  householdExpenses:
-#    totalIterations: 7
-#    inputFields:
-#      householdExpensesType: householdExpensesType
-#      householdExpensesAmount: householdExpensesAmount
-#      householdExpensesFreq: householdExpensesFreq
```
bseeger, line 171: Thanks for the clean up. This feels good!

src/main/resources/templates/mdBenefitsFlow/mailingAddress.html
vrajmohan: @bseeger and I decided to clean this up.

bseeger: Thanks so much for doing this, @vrajmohan! I think it will prevent confusion in the future

bseeger (approved):
Changes look good and functions as discussed

PR #102, opened 2024-03-07
Fix applicant disability PDF mapping

spokenbird:
Fixes the mapping based on the restart of this ticket: https://www.pivotaltracker.com/story/show/187048476

From discussion in said ticket this mapping this will now map "see cover page" for both applicant disability name and applicant pregnancy name in order to be in line with household mappings.


bseeger (approved):
LGTM

PR #106, opened 2024-03-07
Google Drive: send email when material is finished uploading

bseeger:
https://www.pivotaltracker.com/story/show/187120829


src/main/java/org/mdbenefits/app/cli/TransmissionCommands.java
```diff
@@ -34,24 +38,42 @@ public class TransmissionCommands {
     @Value("${google.drive.queen-annes-county-directory-id}")
     private String QUEENANNES_COUNTY_GOOGLE_DIR_ID;
 
+    private List<String> emailRecipients_Baltimore =
+            List.of(
+                    "bseeger+balt@codeforamerica.org",
```
vrajmohan, line 43: These may or may not be considered PII, we should verify.

src/main/resources/messages.properties
```diff
@@ -928,3 +928,7 @@ email-to-recipient.subject=Application Received
 email-to-recipient.body=<html><body><p>We received your Maryland Benefits application. Because you applied for food and/or cash programs, \
   you are automatically considered for an energy benefit (OHEP).</p><p>Application status: In review</p><p>Your county will contact you \
   via mailing address and/or phone number with next steps.</p><p>If you want an update on your case, please call your county.</p></body></html>
+
+email-to-caseworkers.subject=new application case #{0}
+email-to-caseworkers.body=<html><body><p>A new benefit application has been submitted. Please go to the following link to retrieve the application of case number #{0} <a target="_blank" href="{1}">{1}</a></p> \
```
vrajmohan, line 933: It should be confirmation number - case number has a distinct meaning.

src/main/java/org/mdbenefits/app/config/TransmissionConfiguration.java
```diff
@@ -0,0 +1,13 @@
...
+
+import java.util.Map;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@ConfigurationProperties(prefix = "transmission")
+public class TransmissionConfiguration {
+
+    Map<String, String> googleDriveDirectoryId;
```
bseeger, line 11: @spokenbird I'm not really sure what value this config class adds. I know I'm not tracking the variables correctly anymore and things still seem to work fine w/o this functioning.  (I think).  Why do I need this class?

src/main/java/org/mdbenefits/app/cli/TransmissionCommands.java
```diff
@@ -171,12 +190,40 @@ private void transmitBatch(List<Transmission> transmissions) {
...
 
-            // TODO - send emails to state (another ticket)
+            sendEmailToCaseworkers(transmission, confirmationNumber, emailRecipients, newFolder.getUrl(), errorMap);
 
             updateTransmissionStatus(transmission, TransmissionStatus.COMPLETED, errorMap, true);
         });
     }
 
+    // email addresses is a string of email addresses with a space in between
+    private void sendEmailToCaseworkers(Transmission transmission, String confirmationNumber, String emailAddresses,
```
spokenbird, line 200: This has email address as the third parameter but above it's passed as newFolder.gerUrl()?

src/main/java/org/mdbenefits/app/config/TransmissionConfiguration.java
spokenbird: My understanding is it sets the prefix as valid in application.yaml and enables autocompletion. Without it, the application.yaml won't register the values you add there. Just tried removing it and got:

`Could not resolve placeholder 'BALTIMORE_COUNTY_EMAIL_ADDRESSES' in value "${BALTIMORE_COUNTY_EMAIL_ADDRESSES}"`

spokenbird: It was allowing us to have multi keys before a field like foo.bar.thingWeAdd=value, but in this case if it's just `transmission` then I don't think we need it.

spokenbird:
Looks good to me pending the one comment.

src/main/java/org/mdbenefits/app/cli/TransmissionCommands.java
bseeger: Line 193 looks correct to me.  It's sending `emailRecipients` as the third parameter.

PR #114, opened 2024-03-12
At @Scheduled logic to transmission code

bseeger:
Addresses https://www.pivotaltracker.com/story/show/187125884

This will be easier to review once https://github.com/codeforamerica/mdpilot/pull/106 is in, as it builds off of that work.


spokenbird:
Looks good to me. Manually checked and saw transmissions getting sent and files making it to GDrive!

PR #115, opened 2024-03-12
Make demo be a manual deploy, versus automatic

bseeger:
If demo is mean to track prod, we need to make sure it doesn't update automatically and only when we are ready to demo a new feature/change.


spokenbird (approved):
Looks good. We will manually deploy to demo :)

PR #122, opened 2024-03-18
fix 187216187 by navigating to additional household income in flows config

stabai:
Fixes https://www.pivotaltracker.com/n/projects/2688497/stories/187216187

Not sure if there is additional work needed for config, tests, etc.


analoo (approved):
LGTM!

PR #123, opened 2024-03-18
Revamp additional income screen 187100623

bseeger:
Addresses: https://www.pivotaltracker.com/story/show/187100623

This updates the additional income screens to match the design.


src/main/java/org/mdbenefits/app/inputs/MdBenefitsFlow.java
```diff
@@ -204,6 +204,27 @@ public class MdBenefitsFlow extends FlowInputs {
     @NotEmpty(message = "{error.missing-general}")
     private List<String> additionalIncome;
 
+    @NotEmpty(message = "{error.missing-general}")
```
analoo, line 207: This is clever.
1. The first time that the field appears it will be in the additional income page and the value is set to the name. If I don't select the checkbox then the input will not be submitted so this error message will not be triggered.

2. The second time, only fields that were selected in the previous page will have corresponding inputs so if I try to submit a value on the second page, the value will be empty so it will not let me move forward.

src/main/resources/flows-config.yaml
```diff
@@ -215,8 +215,10 @@ flow:
       - name: householdAdditionalIncome
   householdAdditionalIncome:
     nextScreens:
-      - name: householdMoneyOnHand
-        condition: ProgramsIncludeSnap
+      - name: householdAdditionalIncomeAmount
+  householdAdditionalIncomeAmount:
+    beforeSaveAction: ClearUnusedAdditionalIncomeFields
+    nextScreens:
       - name: expensesSignPost
```
analoo, line 222:
```suggestion
      - name: householdMoneyOnHand
```

Small change but I believe that the screen after householdAdditionalIncomeAmount is the householdMoneyOnHand page.

src/main/resources/templates/mdBenefitsFlow/householdAdditionalIncomeAmount.html
```diff
@@ -0,0 +1,126 @@
...
+      <div th:replace="~{fragments/goBack :: goBackLink}"></div>
+      <main id="content" role="main" class="form-card spacing-above-35">
+        <th:block th:replace="~{fragments/icons :: assetsCashAccounts}"></th:block>
+        <th:block
+            th:replace="~{fragments/cardHeader :: cardHeader(header=#{additional-income.amount.header})}"/>
+        <th:block
+            th:replace="~{fragments/form :: form(action=${formAction}, content=~{::formDetails})}">
+          <th:block th:ref="formDetails">
+            <div class="form-card__content">
+              <div th:with="income=${inputData.get('additionalIncome[]')},
```
analoo, line 18: I haven't fully flushed this out fully:

How would you feel about adding a method in the utilities file that returns only the list of Additional Income Type that is relevant in this page.

Then you could loop through that list:
```
                  <th:block
                      th:replace="~{fragments/inputs/money :: money(
                          inputName=${addIncType.getInputName().toString()},
                          label=${#messages.msg(addIncType.getLabelSrc())},
                          placeholder=${0.00}
                  )}"/>
                  ```

analoo (commented):
Looks good, biggest piece of feedback is around the household additional income amount.

src/main/resources/flows-config.yaml
bseeger: Sure, I'll tweak this.

src/main/resources/templates/mdBenefitsFlow/householdAdditionalIncomeAmount.html
bseeger: Thanks for catching this - I cleaned the page up.

src/main/resources/flows-config.yaml
bseeger: I change this and also added a condition for whether or not to show the amounts page.

analoo (approved):
LGTM - Tested it locally and it works as expected.

Tested:
a few items selected
all items selected
no items selected

and confirmed that the PDF was filled out properly

PR #126, opened 2024-03-18
Fix 187194089 by updating confirmation email body

stabai:
I updated the HTML for the email and also had to add the confirmation number to the args.

I added a test that ensures the confirmation number is present in the email body too.

### DIFF
--- a/src/main/java/org/mdbenefits/app/cli/TransmissionCommands.java
+++ b/src/main/java/org/mdbenefits/app/cli/TransmissionCommands.java
@@ -1,6 +1,7 @@
 package org.mdbenefits.app.cli;
 
 import com.amazonaws.services.s3.model.AmazonS3Exception;
+import com.google.api.services.drive.model.File;
 import com.mailgun.model.message.MessageResponse;
 import formflow.library.data.Submission;
 import formflow.library.data.UserFile;
@@ -9,13 +10,6 @@ import formflow.library.email.MailgunEmailClient;
 import formflow.library.file.CloudFile;
 import formflow.library.file.CloudFileRepository;
 import formflow.library.pdf.PdfService;
-import java.io.IOException;
-import java.time.OffsetDateTime;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
-import java.util.concurrent.TimeUnit;
 import lombok.extern.slf4j.Slf4j;
 import org.mdbenefits.app.data.Transmission;
 import org.mdbenefits.app.data.TransmissionRepository;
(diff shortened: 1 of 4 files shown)


src/main/java/org/mdbenefits/app/cli/TransmissionCommands.java
```diff
@@ -27,7 +21,14 @@
 import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.shell.standard.ShellComponent;
 import org.springframework.stereotype.Service;
-import com.google.api.services.drive.model.File;
+
+import java.io.IOException;
```
bseeger, line 25: Generally we try not to change files we are not working in.  Did the style code do this automatically, or is this a preference?   I've never pulled out the java.* before. Is that best practice? 

While I'm thinking of it, we definitely don't do wildcards - that's just an FYI.

src/main/resources/messages.properties
```diff
@@ -936,10 +936,41 @@ doc-upload.expired=The time window to upload documents has run out.
...
+  <p> \
+    If you want an update on your case, please call \
+    <br/>\
+    Queen Anne's County: 401-758-8000 \
+    <br/>\
+    Baltimore County: 1-800-332-6347 \
+  </p> \
+  <p> \
+    <strong>Sending additional documents:</strong>\
+    If you would like to email additional documents, you can email Baltimore County at baltimorecounty.dmc@maryland.gov  or Queen Anne's County at qacfia.customeraccount@maryland.gov. \
```
bseeger, line 961: The apostrophe in Queen Anne's county is not showing up in the email for some reason.  It might be an encoding issue we have sometimes when we copy content from different places (especially figma). If you delete and re-add it, it might solve the issue.

bseeger:
This looks great and works as promised.  Just have a few minor comments

src/main/resources/messages.properties
```diff
@@ -936,10 +936,41 @@ doc-upload.expired=The time window to upload documents has run out.
...
+    <strong>Application Status:</strong>\
+    In review \
+    <br/>\
+    <strong>Next Steps:</strong>\
+    Your county will contact you via email, mail or phone with next steps. \
+  </p> \
+  <p> \
+    If you want an update on your case, please call \
+    <br/>\
+    Queen Anne's County: 401-758-8000 \
```
bseeger, line 955: This apostrophe also doesn't show up (see comment below)

src/main/resources/messages.properties
stabai: Done? How did you verify that?

bseeger (approved):
Looks good - thanks for the changes.

PR #129, opened 2024-03-19
Implement money on hand page

analoo:
This PR :
- Changes the Money on Hand Options
- Uses MoneyOnHand Type to dynamically create MoneyOnHand Options
- Adds some test coverage
- Preparer Logic
- PDF Mapping
- Removal of Old Code
- PDF Field Name updates

https://www.pivotaltracker.com/story/show/187176044


src/main/java/org/mdbenefits/app/data/enums/MoneyOnHandType.java
```diff
@@ -0,0 +1,36 @@
...
+            MAP_BY_VALUE.put(type.value, type);
+        }
+    }
+
+    MoneyOnHandType(String value, String labelSrc) {
+        this.value = value;
+        this.labelSrc = labelSrc;
+    }
+
+    public static String getLabelSrc(String value) {
```
bseeger, line 32: When we talked I mentioned that this may not need to be static. I was confused because it looks like a getter for the field `labelSrc`.  But really this is doing a look up by value.  Could you rename this to be something like `lookupLabelSrcByValue` or, have it return the enum instead, and then pull the label src from it?  I don't have a preference for which you choose, just that this method name be a bit clearer and not collide with the getter name.

analoo: @bseeger - I realize I don't need this method because there seems to be built in getters for value and labelSrc. So I removed the method altogether and confirmed that the code still works as expected.

src/test/java/org/mdbenefits/app/journeys/MdBenefitsFlowJourneyTest.java
```diff
@@ -669,6 +690,27 @@ void completeFlowWithHousehold() {
         testPage.clickButton("I'm done");
 
         assertThat(testPage.getTitle()).isEqualTo(message("seasonal-farmworker.title"));
+        testPage.clickButton("No");
+
```
bseeger, line 694: Thanks for adding these in.

src/main/resources/templates/mdBenefitsFlow/householdMoneyOnHand.html
```diff
@@ -22,27 +21,12 @@
                 content=~{::moneyOnHand})}">
                 <th:block th:ref="moneyOnHand">
                   <th:block
-                      th:replace="~{fragments/inputs/checkboxInSet :: checkboxInSet(inputName='moneyOnHandTypes',value=#{moneyonhand-types.option1}, label=#{moneyonhand-types.option1})}"/>
+                      th:each="option : ${T(org.mdbenefits.app.data.enums.MoneyOnHandType).values()}">
+                    <th:block
+                        th:replace="~{fragments/inputs/checkboxInSet :: checkboxInSet(inputName='moneyOnHandTypes',value=${option.getValue()}, label=#{${option.getLabelSrc()}})}"/>
```
bseeger, line 26: This looks great - I think there are a few variables in the inputs file that can come out, because the second page has been deleted

src/main/java/org/mdbenefits/app/data/enums/MoneyOnHandType.java
```diff
@@ -0,0 +1,23 @@
...
+@Getter
+public enum MoneyOnHandType {
+    CHECKING("checking", "moneyonhand-types.checking-account"),
+    SAVINGS("savings", "moneyonhand-types.savings-account"),
+    BONDS("bonds", "moneyonhand-types.bonds"),
+    CASH_ON_HAND("cash", "moneyonhand-types.cash-on-hand"),
+    MUTUAL_FUNDS("mutual funds", "moneyonhand-types.mutual-funds"),
+    STOCKS("stocks", "moneyonhand-types.stocks");
+
+    private final String value;
```
bseeger, line 16: Will this `value` be the value for the PDF file?  You can use the `MoneyOnHandType.CHECKING.name()` for the value in the html.  Nat won't care that it's capitalized and we have a field that duplicates the name of the enum.  Though I could see having this field for the PDF value, or just using the labelSrc for that as well.

src/main/resources/templates/mdBenefitsFlow/householdMoneyOnHand.html
bseeger: We don't really need to use the enum's `value` field here (we may not even need the enum to have `value`).  You can just use `option.name()` - which will give you the enum name (SAVINGS, CHECKING, etc).   It's easy to look up then when filling in the pdf.  And it's a consistent value.

src/main/resources/templates/mdBenefitsFlow/householdMoneyOnHand.html
```diff
@@ -22,27 +21,12 @@
...
-                  <th:block
-                      th:replace="~{fragments/inputs/checkboxInSet :: checkboxInSet(inputName='moneyOnHandTypes',value=#{moneyonhand-types.option8}, label=#{moneyonhand-types.option8})}"/>
-                  <th:block
-                      th:replace="~{fragments/inputs/checkboxInSet :: checkboxInSet(inputName='moneyOnHandTypes',value=#{moneyonhand-types.option9}, label=#{moneyonhand-types.option9})}"/>
-                  <th:block
-                      th:replace="~{fragments/inputs/checkboxInSet :: checkboxInSet(inputName='moneyOnHandTypes',value=#{moneyonhand-types.option10}, label=#{moneyonhand-types.option10})}"/>
-                  <th:block
-                      th:replace="~{fragments/inputs/checkboxInSet :: checkboxInSet(inputName='moneyOnHandTypes',value='None', label=#{moneyonhand-types.option11}, noneOfTheAbove=true)}"/>
+                      th:replace="~{fragments/inputs/checkboxInSet :: checkboxInSet(inputName='moneyOnHandTypes',value='None', label=#{general.none-of-the-above}, noneOfTheAbove=true)}"/>
```
bseeger, line 29: I used "NONE" for the value in the additional income pages. "None" is fine here, but we should be maybe try to pick one. And we should make sure our alg that checks them compares with no case.

src/main/java/org/mdbenefits/app/data/enums/MoneyOnHandType.java
analoo: Yes - the value here will be doing into the PDF file.

src/main/resources/templates/mdBenefitsFlow/householdMoneyOnHand.html
analoo: I am down for using .name() but want to keep value in for now, as I plan to use that to fill out the PDF

src/main/resources/templates/mdBenefitsFlow/householdMoneyOnHand.html
analoo: Yes. We should definitely pick one. 

I am happy to go with whichever option. In the basic array condition check we use "None" but it's easy to change: https://github.com/codeforamerica/mdpilot/blob/af5064339c1ba23ba092c1e32db42875b5934681/src/main/java/org/mdbenefits/app/submission/conditions/BasicArrayCondition.java#L20

I will use a message value instead of a string in this case and set it to NONE for consistency with your recent code.

analoo: Yes - By this, do you mean that we should have a constant that we use to both define "None/NONE"?

stabai: Sorry, I think I got confused between this comment and the other one that was about an enum. This one is just a string, not an enum, so my suggestion makes no sense.

I feel like trying to describe the way I would write this would just be confusing. Can I send a suggested code change after you merge this?

src/main/resources/templates/mdBenefitsFlow/householdMoneyOnHand.html
```diff
@@ -22,27 +21,12 @@
...
-                  <th:block
-                      th:replace="~{fragments/inputs/checkboxInSet :: checkboxInSet(inputName='moneyOnHandTypes',value=#{moneyonhand-types.option8}, label=#{moneyonhand-types.option8})}"/>
-                  <th:block
-                      th:replace="~{fragments/inputs/checkboxInSet :: checkboxInSet(inputName='moneyOnHandTypes',value=#{moneyonhand-types.option9}, label=#{moneyonhand-types.option9})}"/>
-                  <th:block
-                      th:replace="~{fragments/inputs/checkboxInSet :: checkboxInSet(inputName='moneyOnHandTypes',value=#{moneyonhand-types.option10}, label=#{moneyonhand-types.option10})}"/>
-                  <th:block
-                      th:replace="~{fragments/inputs/checkboxInSet :: checkboxInSet(inputName='moneyOnHandTypes',value='None', label=#{moneyonhand-types.option11}, noneOfTheAbove=true)}"/>
+                      th:replace="~{fragments/inputs/checkboxInSet :: checkboxInSet(inputName='moneyOnHandTypes',value=#{general.none}, label=#{general.none-of-the-above}, noneOfTheAbove=true)}"/>
```
bseeger, line 29: I think a string const would be more effective for the "NONE" value.  Using messages, while fine for our setup since we are only planning on English, would be tricky in a system with multiple languages.  Since this is for our internal logic usage, it would be tricky if it matched the clients language and would make our work harder.    Let's talk about this after standup (with just us devs) and plot a path forward.

bseeger (commented):
LGTM!

PR #131, opened 2024-03-19
Update PDF and PDF map to reflect new fields added

analoo:
This ticket only updates the PDF (and the PDF map for visibility into the new fields created). Below is a screenshot of what the text looks like.

https://www.pivotaltracker.com/story/show/187176044

Screenshot 2024-03-19 at 1 58 04 PM


analoo:
Combined work with other PR

PR #132, opened 2024-03-19
Remove household total income from income summary screen

stabai:
Fixes https://www.pivotaltracker.com/n/projects/2688497/stories/187222921

Removes the total from the html template, and deletes the now-unused message for that section


analoo (approved):
LGTM

PR #133, opened 2024-03-20
Update expenses sign post copy

analoo:
Update to copy so that it matches design:
- removes step
- updates header copy
https://www.pivotaltracker.com/story/show/187223165


bseeger (approved):
LGTM.

bseeger (approved):

PR #134, opened 2024-03-20
Make it so a Google Shared (Team) Drive will work with our code as well. 

bseeger:
Addresses: https://www.pivotaltracker.com/story/show/187213732

 * Moved from using "delete" to "trash" as certain google drive users cannot delete things.
 * Adds more comments to methods for clarity's sake.
 * Adds a "demo" application file so we can differentiate which system email is coming from.

### DIFF
--- a/src/main/java/org/mdbenefits/app/cli/GoogleDriveClient.java
+++ b/src/main/java/org/mdbenefits/app/cli/GoogleDriveClient.java
@@ -39,7 +39,8 @@ public interface GoogleDriveClient {
     List<File> findDirectory(String name, String parentId);
 
     /**
-     * Deletes the specified directory with the matching directory ID.
+     * Hard deletes the specified directory with the matching directory ID; this bypasses the trash feature. Only "Managers" in a
+     * Google Drive team environment can use this call. A "Content Manager" will not have sufficient permissions to use this.
      *
      * @param name        Name of the directory to delete
      * @param directoryId ID of directory to delete
@@ -47,4 +48,14 @@ public interface GoogleDriveClient {
      * @return boolean value indicating success or failure to delete
      */
     boolean deleteDirectory(String name, String directoryId, Map<String, String> errors);
+
+    /**
+     * Puts a specified directory into the trash. A Google Drive user with "Content Manager" role can use this method.
+     *
+     * @param name        Name of the directory to delete
+     * @param directoryId ID of directory to delete
+     * @param errors      error map for any errors to be added to
+     * @return boolean value indicating success or failure to delete
+     */
+    boolean trashDirectory(String name, String directoryId, Map<String, String> errors);
 }
--- a/src/main/java/org/mdbenefits/app/cli/GoogleDriveClient.java
+++ b/src/main/java/org/mdbenefits/app/cli/GoogleDriveClient.java
@@ -46,6 +46,7 @@ public interface GoogleDriveClient {
      * @param directoryId ID of directory to delete
      * @param errors      error map for any errors to be added to
      * @return boolean value indicating success or failure to delete
+     * @see GoogleDriveClient#trashDirectory(String, String, Map)  trashDirectory
      */
     boolean deleteDirectory(String name, String directoryId, Map<String, String> errors);
 
@@ -56,6 +57,7 @@ public interface GoogleDriveClient {
      * @param directoryId ID of directory to delete
      * @param errors      error map for any errors to be added to
      * @return boolean value indicating success or failure to delete
+     * @see GoogleDriveClient#deleteDirectory(String, String, Map)  deleteDirectory
      */
     boolean trashDirectory(String name, String directoryId, Map<String, String> errors);
 }
--- a/src/main/java/org/mdbenefits/app/cli/GoogleDriveClientImpl.java
+++ b/src/main/java/org/mdbenefits/app/cli/GoogleDriveClientImpl.java
@@ -6,6 +6,7 @@ import com.google.api.client.googleapis.media.MediaHttpUploader;
 import com.google.api.client.http.ByteArrayContent;
 import com.google.api.client.http.HttpRequestInitializer;
 import com.google.api.client.json.gson.GsonFactory;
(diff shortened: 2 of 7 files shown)


src/main/java/org/mdbenefits/app/cli/GoogleDriveClientImpl.java
```diff
@@ -33,7 +36,11 @@ public class GoogleDriveClientImpl implements GoogleDriveClient {
     private final String GOOGLE_CREDS = System.getenv("GOOGLE_DRIVE_CREDS");
```
bseeger, line 36: So this works, but I'm debating if it should be something we pull in this way, or should be something we pull in via the appliation.yaml setup, like we do most other things.

src/main/java/org/mdbenefits/app/cli/GoogleDriveClient.java
```diff
@@ -39,12 +39,23 @@ public interface GoogleDriveClient {
     List<File> findDirectory(String name, String parentId);
 
     /**
-     * Deletes the specified directory with the matching directory ID.
+     * Hard deletes the specified directory with the matching directory ID; this bypasses the trash feature. Only "Managers" in a
+     * Google Drive team environment can use this call. A "Content Manager" will not have sufficient permissions to use this.
      *
```
stabai, line 44: non-blocking: Consider adding a "see also" that references the `trashDirectory` method.

src/main/java/org/mdbenefits/app/cli/GoogleDriveClientImpl.java
stabai: That seems like a question beyond the scope of this PR. I'm not particularly used to Tomcat's configuration, but I expect to see things about serving there (e.g. timeouts, file size limits, etc) rather than the application configuration for the service itself.

src/main/java/org/mdbenefits/app/cli/GoogleDriveClientImpl.java
```diff
@@ -33,7 +36,11 @@ public class GoogleDriveClientImpl implements GoogleDriveClient {
...
     private final Drive service;
 
-    public GoogleDriveClientImpl() throws IOException, GeneralSecurityException {
+    private final String teamDriveId;
+
+    public GoogleDriveClientImpl(@Value("${transmission.google-drive-directory-id.team-drive:''}") String teamDriveId)
```
stabai, line 41: For my own education: I'm curious what the argument to `@Value` does.

src/main/java/org/mdbenefits/app/cli/GoogleDriveClientImpl.java
```diff
@@ -124,11 +149,33 @@ public List<File> findDirectory(String name, String parentId) {
...
 
+    @Override
+    public boolean trashDirectory(String name, String directoryId, Map<String, String> errors) {
+        try {
+            File metadata = new File();
+            metadata.setTrashed(true);
+            service.files()
+                    .update(directoryId, metadata)
+                    .setSupportsTeamDrives(true)
+                    .setSupportsAllDrives(true)
```
stabai, line 160: It seems unobvious that we would set these when trashing a directory. Can you add a comment about why this is necessary?

src/main/java/org/mdbenefits/app/cli/GoogleDriveClientImpl.java
```diff
@@ -145,6 +192,10 @@ private void listFiles(String folderId, int numberOfFiles) throws IOException {
                 .setQ(String.format("'%s' in parents and trashed = false", folderId))
                 .setSpaces("drive")
                 .setPageSize(numberOfFiles)
+                .setDriveId(teamDriveId)
+                .setCorpora("drive")
+                .setSupportsAllDrives(true)
+                .setSupportsTeamDrives(true)
```
stabai, line 198: Same

src/main/java/org/mdbenefits/app/cli/TransmissionCommands.java
```diff
@@ -203,7 +217,16 @@ private void transmitBatch(List<Transmission> transmissions) {
...
     }
 
-    // email addresses is a string of email addresses with a space in between
+    /**
+     * Send email about the transmission to specified email addresses.
+     *
+     * @param transmission       the transmission we are concerned with
+     * @param confirmationNumber the confirmation number of the application
+     * @param emailAddresses     the email addresses to send it to - these should be a string of email addresses with a space in
+     *                           between addresses
```
stabai, line 226: non-blocking: Consider giving this API parameter a more semantically-accurate type (`List<String>`)

stabai (approved):

src/main/java/org/mdbenefits/app/cli/GoogleDriveClientImpl.java
bseeger: @stabai the argument is a yaml field name that is located in the application.yaml (& profile overrides) file. It will pull from that field, if set in the yaml file, or default to whatever follows the colon, if supplied.

src/main/java/org/mdbenefits/app/cli/GoogleDriveClientImpl.java
bseeger: I need to double check that this is necessary here. The only reason I include these is that when we are accessing prod it's a Team drive and requires these be set.  The staging folder is not a team drive and doesn't need these. 

I don't quite know the boundaries of when these are needed and when they are not, so I'll double check it here. 

Does seem to be needed to access data.

bseeger: So it appears that if I don't include this I don't get access to the Team drive environment.  Even in the context of trashing a directory.

src/main/java/org/mdbenefits/app/cli/TransmissionCommands.java
bseeger: Noting this for future work.
https://www.pivotaltracker.com/story/show/187296734

PR #138, opened 2024-03-21
Refactor none of the above handling for checkbox lists

stabai:


bseeger:
This looks great and works as promised.  There are a few other spots that could benefit from this change: 

https://github.com/search?q=repo%3Acodeforamerica%2Fmdpilot%20noneOfTheAbove&type=code

Some of them you have addressed and at least one spot will need the "none" option teased out.

stabai:
> ~This looks great and works as promised. There are a few other spots that could benefit from this change:~
> 
> ~https://github.com/search?q=repo%3Acodeforamerica%2Fmdpilot%20noneOfTheAbove&type=code~
> 
> ~Some of them you have addressed and at least one spot will need the "none" option teased out.~
> 
> Actually, I'd only have you address the SelectApplication one as Ana is working on the expenses ones (household expenses / utilities / medical expenses). Race and Ethnicity should probably stay as is, as "other" is a more appropriate value there.

I'm not sure I know exactly how we'd want to approach it for that one. The [enum value](https://github.com/codeforamerica/mdpilot/blob/817cea7145f82eac5def8cc0d412d31dae2a0ed8/src/main/java/org/mdbenefits/app/data/enums/ApplicantObjective.java#L14) of `OTHER` seems appropriate there to me. Do you think the value should change to `NONE`, or just that the same logic for identifying a "none of the above" situation should be applied there (with `OTHER` being the key phrase to search for)?

bseeger:
> I'm not sure I know exactly how we'd want to approach it for that one. The [enum value](https://github.com/codeforamerica/mdpilot/blob/817cea7145f82eac5def8cc0d412d31dae2a0ed8/src/main/java/org/mdbenefits/app/data/enums/ApplicantObjective.java#L14) of OTHER seems appropriate there to me. Do you think the value should change to NONE, or just that the same logic for identifying a "none of the above" situation should be applied there (with OTHER being the key phrase to search for)?

Good point. Maybe we should just leave this PR as is and leave that page as is.  Other does seem more appropriate there rather than none.

bseeger (approved):
LGTM

PR #146, opened 2024-03-25
Add help accordion to household income screen and populate in PDF

stabai:
Fixes https://www.pivotaltracker.com/n/projects/2688497/stories/187216213


src/main/java/org/mdbenefits/app/preparers/IncomeDetailsPreparer.java
```diff
@@ -39,28 +39,45 @@ public Map<String, SubmissionField> prepareSubmissionFields(Submission submissio
...
+        boolean hasIncome = income != null && !income.isEmpty();
+        fields.put(
+                "householdHasEarnedIncome",
+                new SingleField("householdHasEarnedIncome", hasIncome ? "true" : "false", null));
 
+        if (income != null) {
+            int i = 1;
+            for (Map<String, Object> incomeDetails : income) {
                 var employeeName = incomeDetails.get("householdMemberJobAdd");
+                if (employeeName.equals("you")) {
```
bseeger, line 54: I can't recall if we stored the applicant name as "you" before or if we translated it in an action after the page.  @analoo do you know? Given that you have to add this logic, perhaps we already had changed "you" -> "firstname, lastname" before we got this far.  There are some pieces of code that make me think we leave it as "you", so perhaps this is just fine.

bseeger (approved):
LGTM - One question to figure out, but if you and @analoo feel good about it, feel free to merge.

src/main/java/org/mdbenefits/app/preparers/IncomeDetailsPreparer.java
stabai: It was filling the PDF in as "you" before, which was why I added this if statement.

analoo: From what I recall we use 'you' to refer to the candidate so this update makes sense to me.

PR #148, opened 2024-03-25
Hide the expedited snap notification page.

bseeger:
Addresses the 2nd part of: https://www.pivotaltracker.com/story/show/187270727

The boxes are already out of the PDF file. That was done in a separate [PR](https://github.com/codeforamerica/mdpilot/pull/145)


bseeger:
Just an FYI that we plan to add expedited snap back in, hence commenting out and not deleting it.

PR #150, opened 2024-03-25
Revamp medical expenses page to match design. (front end, no PDF)

bseeger:


src/main/java/org/mdbenefits/app/submission/conditions/HasHouseholdMemberDisabledOrOver60.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.Map;
+import org.springframework.stereotype.Component;
+
+@Component
+public class HasHouseholdMemberDisabledOrOver60 implements Condition {
+
+    @Override
+    public Boolean run(Submission submission) {
+        var inputData = submission.getInputData();
+        List<Map<String, Object>> householdMembers = (List) inputData.getOrDefault("household", emptyList());
```
analoo, line 18: @bseeger - added an empty list as default because this was breaking when household does not exist.

src/main/java/org/mdbenefits/app/submission/conditions/HasMedicalExpenses.java
```diff
@@ -8,6 +8,6 @@ public class HasMedicalExpenses extends BasicArrayCondition {
 
     @Override
     public Boolean run(Submission submission) {
-        return checkHasValidSelection(submission, "householdMedicalExpenses[]");
+        return checkHasValidSelection(submission, "medicalExpenses[]");
```
analoo, line 11: this was the missing piece.

bseeger: nice catch! I kept running into issues around the name change.

PR #151, opened 2024-03-26
Expenses pdf changes

analoo:


src/main/resources/pdf-map.yaml
```diff
@@ -200,6 +209,33 @@ inputFields:
   homeExpenseWaterFrequency: expense_water_frequency
   homeExpenseWoodOrCoalFrequency: expense_wood_frequency
 
+  expenseSection8Housing: expense_section_8_bool
+  expenseHeatIncludedRent: expense_heat_included_rent_bool
```
bseeger, line 213: I'm noticing these don't show up automatically, but I suspect it's because they are lists in the data: 

```
"expenseSection8Housing[]": [
    "true"
  ],
  "expenseHeatIncludedRent[]": [
    "true"
  ],
```

bseeger:
Looking really good!

src/main/resources/pdf-map.yaml
analoo: ah! This is the issue I was coming up on! I couldn't figure out why these checkmarks were not showing up

src/test/java/org/mdbenefits/app/preparers/HouseholdExpensesPreparerTest.java
```diff
@@ -46,4 +46,11 @@ public void shouldAddFieldsForHouseholdExpensesFrequency() {
         assertThat(result.get("homeExpensePhoneFrequency"))
             .isEqualTo(new SingleField("homeExpensePhoneFrequency", "Monthly", null));
     }
+
+    private Map<String, SubmissionField> withNoMedicalExpenses(Map<String, SubmissionField> results) {
+        assertThat(results.keySet()).contains("hasMedicalExpenses");
```
bseeger, line 51: This method is doing double duty - it's removing medical expenses and asserting that they are there and then gone.  

I think it should only do one thing -- remove medical expenses if they are there.  This assumes they are there and asserts if not.  

I would separate these concerns - if this method is meant to strip out medical expenses, if there, then have it just do that and not assert in the process. I think the assert in a method that hands something back to later assert seems way too complicated to me.

bseeger (approved):
Looks good. There's one spot in test that can be changed a little. 

Once you're satisfied the PDF is working, feel free to merge.

PR #155, opened 2024-03-26
Update confirmation page.

bseeger:
addresses: https://www.pivotaltracker.com/story/show/187056084
This PR: 
* introduces a redirect after 5 seconds on the very last page (you have to submit feedback or click “I’m done”).   Exercising that would be good to ensure that it redirects successfully and clears the session.   I've tested with Brave and Chrome and things look good there. 
* introduces logic that will cause the pdf download link to be suppressed if we are in Prod.


src/test/java/org/mdbenefits/app/testutils/Page.java
```diff
@@ -47,7 +47,7 @@ public String getHeader() {
 
     public void goBack() {
         await().atMost(20, TimeUnit.SECONDS).until(
-            () -> !(driver.findElements(By.id("back-link")).isEmpty())
+                () -> !(driver.findElements(By.id("back-link")).isEmpty())
```
bseeger, line 50: grumble, grumble, much of this file is just syntax formatting. I'll mark where there's an actual change.

src/test/java/org/mdbenefits/app/testutils/Page.java
```diff
@@ -339,15 +339,18 @@ public WebElement findAccordionByButtonText(String buttonText) {
                 .orElse(null);
     }
 
-    public boolean elementDoesNotExistById(String id) {
+    public boolean elementExistsById(String id) {
```
bseeger, line 342: I created this method and revamped the one below it so they played off of each other in a logical fashion.

src/main/resources/templates/mdBenefitsFlow/howThisWorks.html
```diff
@@ -24,10 +24,10 @@
         <th:block th:replace="~{fragments/honeycrisp/accordion :: accordion(
           buttonLabel=#{how-this-works.who-can},
           content=~{::who})}">
-            <th:block th:ref="who">
-             <p th:text="#{how-this-works.who-can-l1}"></p>
-              <p th:text="#{how-this-works.who-can-l2}"></p>
-            </th:block>
+          <th:block th:ref="who">
```
bseeger, line 27: I was going to edit something in this file and then changed my mind. But it did some auto formatting in the mean time.

src/main/resources/flows-config.yaml
```diff
@@ -299,11 +299,15 @@ flow:
   confirmation:
     beforeDisplayAction: FormatSubmittedAtDate
     nextScreens:
-      - name: confirmation
+      - name: finalPage
+  finalPage:
+    nextScreens:
+      - name: finalPage
```
stabai, line 305: This looks like an endless loop...?

bseeger: Yeah, it should be okay though, as the finalPage doesn't have any way to "submit". It will redirect to the index page after 5 seconds.

PR #156, opened 2024-03-26
Remove hourly job flow and update PDF prep

stabai:
Fixes:
- https://www.pivotaltracker.com/n/projects/2688497/stories/187216234
- https://www.pivotaltracker.com/n/projects/2688497/stories/187216219


src/test/java/org/mdbenefits/app/data/SubmissionTestBuilder.java
```diff
@@ -1,12 +1,12 @@
 package org.mdbenefits.app.data;
 
 import formflow.library.data.Submission;
-import formflow.library.inputs.FieldNameMarkers;
+import org.mdbenefits.app.data.enums.CitizenshipStatus;
```
analoo, line 4: This seems to have accidentally snuck in,

analoo (approved):

src/test/java/org/mdbenefits/app/data/SubmissionTestBuilder.java
stabai: Turned off my "Optimize imports on the fly setting", but it might be good if we used a standardized code format instead so that it just avoids conflicts.

PR #159, opened 2024-03-27
Fix contact info nudge to only pay attention to phone numbers

bseeger:
Addresses changes to:187081382

Snuck in this fix because it kept blowing up here on me:  Fixes a spot where if the `applicantSex` was not set (aka, we skipped around in the flow), it will give a solid default value versus throw an exception.


analoo (approved):
LGTM

PR #162, opened 2024-03-27
Updates few changes

analoo:
Mixed together a few changes:
1. Changed 20 - 25
2. Fixed some small copy issues
3. Added logic to expenses preparer that would set value for dependent care to an empty string if it wasn't included
4. Added logic to append "/month" to child support expenses preparer


src/main/java/org/mdbenefits/app/preparers/HouseholdExpensesPreparer.java
```diff
@@ -59,9 +59,16 @@ public Map<String, SubmissionField> prepareSubmissionFields(Submission submissio
                 var dependentCareExpensesAmount = (String) submission.getInputData().getOrDefault("expensesDependentCare", "");
                 results.put("expensesDependentCare", new SingleField("expensesDependentCare", dependentCareExpensesAmount + " /month", null));
 
+            } else {
+                results.put("hasDependentCareExpenses", new SingleField("hasDependentCareExpenses", "", null));
             }
         }
 
+        var expensesChildSupport = (String) submission.getInputData().getOrDefault("expensesChildSupport", "");
```
bseeger, line 67: Can you note this in the description of the PR as well? Looks like this is adding the PDF fields for the child and dependent expenses.

analoo: Yes. Does this suffice?

Added logic to append "/month" to child support expenses preparer

bseeger: Oh, that was already in the description, wasn't it?  Sorry, was just going by the title and switching context a bit today.

PR #165, opened 2024-03-27
Add ssn section 187323767

analoo:
Adds new citizenship page and move ssn items to it
Removes requirement for ssn but adds a pattern check to ensure that 9 digits are added
writes citizenship status for the applicant to pdf
updates test flow


bseeger (approved):
LGTM

PR #175, opened 2024-03-28
Add privacy policy to rights and responsibility page.

bseeger:
Addresses: https://www.pivotaltracker.com/story/show/187270739

### DIFF
--- a/src/main/resources/templates/fragments/rightsAndResponsibilitiesFragments.html
+++ b/src/main/resources/templates/fragments/rightsAndResponsibilitiesFragments.html
@@ -458,3 +458,10 @@
         satisfactory immigration status.</p>
   </div>
 </th:block>
+<th:block th:fragment="privacyPolicy">
+  <div class="text--small">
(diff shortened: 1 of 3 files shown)


src/main/resources/templates/fragments/rightsAndResponsibilitiesFragments.html
```diff
@@ -458,3 +458,10 @@ <h3 class="text--small">SNAP/EBT Card: Multiple Card Replacements</h3>
         satisfactory immigration status.</p>
   </div>
 </th:block>
+<th:block th:fragment="privacyPolicy">
+  <div class="text--small">
+    <p>I understand that the information I’ve submitted is housed outside of the State of Maryland,
```
spokenbird, line 463: Why not message properties keys here?

bseeger: That's the ideal. Vraj choose to hard code it because there is so much text and I just kept going with that. It is one area that we will have to revisit if we do include other languages.

PR #177, opened 2024-03-28
Update cover page 187325805

analoo:
- Updates disability and pregnancy logic to properly fill in the pdf.
- Most of the work is around combining the logic and adding tests
- Sample PDF attached

[mdBenefitsFlow_69da2d1c-96ae-4b12-bb5a-c6b22d2d2947 (1).pdf](https://github.com/codeforamerica/mdpilot/files/14796267/mdBenefitsFlow_69da2d1c-96ae-4b12-bb5a-c6b22d2d2947.1.pdf)

### DIFF
--- a/src/main/java/org/mdbenefits/app/utils/SubmissionUtilities.java
+++ b/src/main/java/org/mdbenefits/app/utils/SubmissionUtilities.java
@@ -128,6 +128,24 @@ public class SubmissionUtilities {
         return names;
     }
 
+    public static String householdMemberFullNameFormatted(Map<String, Object> householdMember) {
+        String fullName = String.format("%s, %s", householdMember.get("householdMemberLastName"), householdMember.get("householdMemberFirstName"));
+        if (householdMember.get("householdMemberMiddleName") != null) {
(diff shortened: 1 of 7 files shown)


src/main/java/org/mdbenefits/app/utils/SubmissionUtilities.java
```diff
@@ -128,6 +128,24 @@ public static List<String> getHouseholdMemberNames(Submission submission) {
...
+        }
+        return fullName;
+    }
+
+    public static String applicantFullNameFormatted(Submission submission){
+        Map<String, Object> inputData = submission.getInputData();
+
+        String fullName = String.format("%s, %s", inputData.get("lastName"), inputData.get("firstName"));
+        if (inputData.get("middleName") != null) {
+            fullName += ", " + inputData.get("middleName");
```
stabai, line 144: I've never seen a comma between the first and middle name when using this format. It's usually `Jones, James Earl`. That would mean the code should be:

```suggestion
            fullName += " " + inputData.get("middleName");
```

stabai (approved):

PR #178, opened 2024-03-28
Update note about consent for email field

stabai:
Fixes https://www.pivotaltracker.com/n/projects/2688497/stories/187325621

I couldn't get it to put the "Optional" word in its own paragraph because the help text doesn't allow html tags.


analoo (approved):
@stabai LGTM - We could create a whole new fragment that takes in html but it seems overkill. The only other solution I could think of is adding the optional as help text and the additional lines below.

PR #183, opened 2024-03-29
Update the pregnancy and disability name to be first middle last

analoo:


spokenbird (approved):
Over all looks good, wondering if we should add a test case or 2 that include a middle name?

PR #190, opened 2024-04-01
Update deploy-staging.yaml to run after test completion

spokenbird:


.github/workflows/deploy-staging.yaml
```diff
@@ -1,8 +1,10 @@
 name: Deploy to Aptible Staging
 
 on:
-  push:
-    branches: [ main ]
+  workflow_run: 
+    workflows: [ "Run tests" ]
+    types: [ completed ]
+    branches: [ update-deploy-to-staging ]
```
bseeger, line 7: what is this branch? Is it a temporary branch created in this process?

spokenbird: Yep, for testing purposes

bseeger: Oh, you're using this for testing.  Is that right?

bseeger: So for for the final PR, it will say "main"?

spokenbird: It should yes, but I'm realizing this isn't going to work the way I had hoped and now I'm not sure how to test this without merging it to main first 😓

bseeger: Might be fine to merge to main first -- you won't be hurting anything with iterating on this there.  I think just warn us all first so we know that things might fail for a limited amount of time.

PR #197, opened 2024-04-04
Remove amount requirement for inputs 187359678

analoo:
This PR makes the following functionality:
- When a user selects household expenses, medical expenses, or additional incomes
- They are asked to provide a $ value on the follow up page
- The user can move forward as the $ values are not longer required
    - If the user inputs the values of 0, they will not see an error message
    - If the user inputs any value that is not a digit, they will be an error message
    - If the user inputs values that do not match this pattern (X.XX, X) they will see an error message
 - In the PDF, we will see a blank space when the user did not input $ value data


bseeger (approved):
LGTM

PR #199, opened 2024-04-05
Implements Minimum app flow

analoo:
Implements Minimum app Logic

On the confirm contact info page
A user can select "Submit an incomplete application now with only the above information."
And they will be taken to minumAppSubmission page
If a user can select "Yes, I'll answer questions" and they are taken to the householdSignPost page
If a user selects, "No, I'll submit" then they are taken to the legalStuff Page


src/main/resources/templates/mdBenefitsFlow/minimumAppSubmission.html
```diff
@@ -0,0 +1,40 @@
...
+            content=~{::minimumAppContent})}">
+              <th:block th:ref="minimumAppContent">
+                <p th:utext="#{minimum-app-confirmation.expand1}"></p>
+              </th:block>
+          </th:block>
+          <div class="boxed-content spacing-below-60 notice-container">
+            <p th:utext="#{minimum-app-confirmation.p1}"></p>
+          </div>
+        </div>
+        <div class="form-card__footer">
```
analoo, line 29: Add a field that tracks if this isMinimumApp

PR #205, opened 2024-04-10
Add retry count mechanism and virus scan check to google drive transmission

bseeger:
Addresses: https://www.pivotaltracker.com/story/show/187339289 and https://www.pivotaltracker.com/story/show/187212763

Relies on this change getting into form flow library: https://github.com/codeforamerica/form-flow/pull/542


build.gradle
```diff
@@ -44,8 +44,13 @@ dependencies {
...
 
-    implementation "org.codeforamerica.platform:form-flow:${formFlowLibraryVersion}"
-    println "📚Using form flow library ${formFlowLibraryVersion}"
+    if (profile == 'dev' || useLocalLibrary == 'true') {
+        implementation fileTree(dir: "$rootDir/../form-flow/build/libs", include: '*.jar')
+        println "📦 Using local library"
+    } else {
+        implementation "org.codeforamerica.platform:form-flow:${formFlowLibraryVersion}"
+        println "📚Using form flow library ${formFlowLibraryVersion}"
+    }
```
bseeger, line 53: I think this change is temporary and I will reverse this once the FFL changes have been released.

src/main/java/org/mdbenefits/app/cli/TransmissionCommands.java
```diff
@@ -273,22 +302,48 @@ private void handleError(Transmission transmission, String errorKey, String erro
...
         transmission.setErrors(errorMap);
         if (markSent) {
             transmission.setSentAt(OffsetDateTime.now());
         }
+
+        // don't increment retry when just marking as in process
+        if (!status.equals(TransmissionStatus.TRANSMITTING)) {
+            int retryCount = transmission.getRetryCount();
+            retryCount++;
+            transmission.setRetryCount(retryCount);
```
bseeger, line 342: Question: if the retry count is == to the max, should we set status differently, or keep it as QUEUED? It will not get picked up again, but maybe the status should reflect this as well?

bseeger: How should we notify ourselves if an transmission fails to send after 5 tries?  We should put in some code here that causes us to be alerted.  Maybe an ERROR with a specific message and then a Datadog monitor for that message???   @stabai @spokenbird @analoo?

stabai: It's easiest if there's a logging level that always triggers an alert. Do we have something like `CRITICAL` that I can set up to do that?

bseeger: Hmmm... no CRITICAL that I can see.

PR #208, opened 2024-04-11
Remove extra spaces in names 187409488

analoo:


src/main/java/org/mdbenefits/app/submission/actions/ReformatHouseholdNameBeforeSaving.java
```diff
@@ -0,0 +1,30 @@
...
+    public void run(Submission submission, String id) {
+        ArrayList<HashMap<String, Object>> householdSubflow = (ArrayList<HashMap<String, Object>>) submission.getInputData().get("household");
+        if (householdSubflow != null) {
+            HashMap<String, Object> householdMember = householdSubflow.stream()
+                .filter(iteration -> id.equals(iteration.get("uuid"))).toList().get(0);
+            String firstName = (String) householdMember.remove("householdMemberFirstName");
+            String middleName = (String) householdMember.remove("householdMemberMiddleName");
+            String lastName = (String) householdMember.remove("householdMemberLastName");
+
+            householdMember.put("householdMemberFirstName", firstName.trim());
```
bseeger, line 25: I think you can just do a `put` w/o the remove, as it should overwrite the original entry. 
This is good for now, but ultimately, it seems like this is something FFL should do by default. Maybe someday we can get something like this in the library.

src/main/java/org/mdbenefits/app/submission/actions/ReformatHouseholdNameBeforeSaving.java
```diff
@@ -0,0 +1,30 @@
...
+
+@Component
+public class ReformatHouseholdNameBeforeSaving implements Action {
+    @Override
+    public void run(Submission submission, String id) {
+        ArrayList<HashMap<String, Object>> householdSubflow = (ArrayList<HashMap<String, Object>>) submission.getInputData().get("household");
+        if (householdSubflow != null) {
+            HashMap<String, Object> householdMember = householdSubflow.stream()
+                .filter(iteration -> id.equals(iteration.get("uuid"))).toList().get(0);
+            String firstName = (String) householdMember.remove("householdMemberFirstName");
```
analoo, line 21: Currently, empty fields are have an empty string but in the future we should use getOrDefault and only save if the string is not empty.

stabai (approved):

src/test/java/org/mdbenefits/app/submission/actions/ReformatNameBeforeSavingTest.java
stabai, line 1: Love these tests!

PR #218, opened 2024-04-24
Add logic to subflow preparers to skip incomplete records

bseeger:
Also includes minor tweaks to the 9701 to hide expedited and clean it up a little.

We were seeing an error in production where a user had an incomplete income loop, but they hadn't gotten to the point in the loop that cleans out the incomplete iterations. They went into the income subflow, then went back out and skipped it completely.  That left an incomplete iteration in the record, which caused the income preparer to throw an exception.


spokenbird (approved):
Looks good!

PR #222, opened 2024-04-29
Attempt to add (back) in accessibility tests to the code base.

bseeger:
LA and IL have these tests enabled, so this is working to bring them into this code base.

This is not functioning yet. I think there is an issue with the active profile and or tags ("a11y") - ("test" vs "accessibilityTest").  I haven't had time to sort through it yet.


bseeger:
This is no longer needed and @spokenbird will be doing something similar in the form flow library directly.

PR #224, opened 2024-04-29
fix spacing on household details screen

bseeger:
Addresses:  https://www.pivotaltracker.com/story/show/187121194

This is only partially fixed - the link color is changed, but I'm unsure how to make the spans look nice. I suspect it depends on the length of the the text the span is covering. Not sure if we are tying to avoid tables, but a table could lend nice formatting here.


bseeger:
Not sure if this is really necessary anymore.

PR #228, opened 2024-05-03
Make the expedited SNAP flow exposed in the minimal flow

bseeger:
Addresses: #187545545

### DIFF
--- a/src/main/java/org/mdbenefits/app/inputs/MdBenefitsFlow.java
+++ b/src/main/java/org/mdbenefits/app/inputs/MdBenefitsFlow.java
@@ -333,6 +333,9 @@ public class MdBenefitsFlow extends FlowInputs {
     // Expedited Snap Start
     private String isApplyingForExpeditedSnap;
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/mdbenefits/app/inputs/MdBenefitsFlow.java
```diff
@@ -333,6 +333,9 @@ public class MdBenefitsFlow extends FlowInputs {
     // Expedited Snap Start
     private String isApplyingForExpeditedSnap;
 
+    // Minimum App flow
+    private String usedMinimumApplicationFlow;
+
```
bseeger, line 338: Though not asked for, I thought it might be useful to track if they've gone through the minimal application flow. It doesn't mean they didn't go back and fill out the whole app, but it does mean they may have submitted just a minimal app.

spokenbird (approved):

spokenbird:
Looks good!

PR #234, opened 2024-05-10
Add back in expedited flag to PDF and tweak some fields.

bseeger:
Resolves https://www.pivotaltracker.com/story/show/187564802

 * Adds "Likely expedited" box into PDF. 
 * Updates logic for not marking pregnancy and disability when the app is a minimal application
 * Makes sure that the applicant is marked as applying, in the case of a minimum application.


spokenbird (approved):
Tested locally and works as expected, looks good to go!

PR #236, opened 2024-05-10
Update state template to autoselect MD

spokenbird:


bseeger (approved):
LGTM! I think longer term this should go into Form Flow Builder -- we should be able to send in a variable for selecting the state.  Can you make a ticket over there for this?

PR #237, opened 2024-05-10
Put local or maven library selection logic back in build.gradle

spokenbird:


bseeger:
Is this temporary? I think we should always use the library and only do this in our dev env. when we need it.

PR #242, opened 2024-05-16
finishes adding back in expedited snap check to main flow

bseeger:
https://www.pivotaltracker.com/story/show/187564812

### DIFF
--- a/src/main/java/org/mdbenefits/app/inputs/MdBenefitsFlow.java
+++ b/src/main/java/org/mdbenefits/app/inputs/MdBenefitsFlow.java
@@ -192,12 +192,9 @@ public class MdBenefitsFlow extends FlowInputs {
     @NotBlank(message = "{error.missing-pay-period}")
     private String payPeriod;
 
-    // NOTE: this is really pay for the last 30 days
-    // and not the payPeriodAmount!  Don't multiply it out
-    // unless other logic is changed.
     @Money(message = "{error.invalid-money}")
     @NotBlank(message = "{error.missing-dollar-amount}")
-    private String payPeriodAmount;
+    private String payAmountFor30Days;
 
     @NotEmpty(message = "{error.missing-general}")
     private List<String> additionalIncome;
--- a/src/main/java/org/mdbenefits/app/inputs/MdBenefitsFlow.java
+++ b/src/main/java/org/mdbenefits/app/inputs/MdBenefitsFlow.java
@@ -192,6 +192,9 @@ public class MdBenefitsFlow extends FlowInputs {
     @NotBlank(message = "{error.missing-pay-period}")
     private String payPeriod;
 
+    // NOTE: this is really pay for the last 30 days
+    // and not the payPeriodAmount!  Don't multiply it out
+    // unless other logic is changed.
     @Money(message = "{error.invalid-money}")
     @NotBlank(message = "{error.missing-dollar-amount}")
     private String payPeriodAmount;
@@ -349,6 +352,9 @@ public class MdBenefitsFlow extends FlowInputs {
 
     // Expedited Snap Income
     private String incomeLessThan150;
+
+    // Expedited SNAP confirmation Income < Expenses
+    private String clientConfirmsInputLessThanExpenses;
 
     // Household 30 Day Income
     @Money(message = "{error.invalid-money}")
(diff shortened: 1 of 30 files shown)


src/main/java/org/mdbenefits/app/inputs/MdBenefitsFlow.java
```diff
@@ -192,6 +192,9 @@ public class MdBenefitsFlow extends FlowInputs {
     @NotBlank(message = "{error.missing-pay-period}")
     private String payPeriod;
 
+    // NOTE: this is really pay for the last 30 days
+    // and not the payPeriodAmount!  Don't multiply it out
+    // unless other logic is changed.
```
spokenbird, line 197: Should we rename the input?

src/main/java/org/mdbenefits/app/submission/actions/CheckExpeditedSnapEligibility.java
```diff
@@ -21,41 +23,37 @@ public class CheckExpeditedSnapEligibility implements Action {
...
+                .equals("true");
+        boolean isBelowIncomeThreshhold = inputData.getOrDefault("incomeLessThan150", "false").toString().equals("true");
+
+        boolean isEligibleByIncomeAndCashOnHandLessThanExpenses =
+                householdIncomeAmount.add(moneyOnHandAmount).compareTo(expenseCalculator.totalUtilitiesExpenses()) <= 0;
+
+        boolean isEligibleForExpeditedSnap =
+                (isBelowMoneyOnHandThreshhold && isBelowIncomeThreshhold) || (
+                        isMigrantOrSeasonalFarmWorker && isBelowMoneyOnHandThreshhold)
+                        || expenseCalculator.totalUtilitiesExpenses().compareTo(BigDecimal.ZERO) > 0
```
spokenbird, line 54: should the last part of this statement after the final || be wrapped in parens like the others?

src/main/java/org/mdbenefits/app/submission/conditions/HasMoreThan100OnHand.java
```diff
@@ -0,0 +1,18 @@
...
+import java.util.Map;
+import org.springframework.stereotype.Component;
+
+@Component
+public class HasMoreThan100OnHand implements Condition {
+
+    @Override
+    public Boolean run(Submission submission) {
+        Map<String, Object> inputData = submission.getInputData();
+        return inputData.getOrDefault("householdMoneyOnHandLessThan100", "false")
```
spokenbird, line 14: If they didn't answer/see this question, won't this still end up being truthy? Is that desired?

src/main/java/org/mdbenefits/app/utils/ExpenseCalculator.java
```diff
@@ -0,0 +1,52 @@
...
+
+@Slf4j
+public class ExpenseCalculator {
+
+    Submission submission;
+
+    public ExpenseCalculator(Submission submission) {
+        this.submission = submission;
+    }
+/*
```
spokenbird, line 22: Just noting we should remove the commented code.

src/main/java/org/mdbenefits/app/utils/IncomeCalculator.java
```diff
@@ -50,5 +53,6 @@ public static double futureIncomeForJob(Map<String, Object> job) throws NumberFo
             // based on 30D estimate
             return payPeriodAmount;
         }
+ */
```
spokenbird, line 56: Also remove commented out code?

src/test/java/org/mdbenefits/app/journeys/MdBenefitsFlowJourneyTest.java
```diff
@@ -535,9 +659,9 @@ void completeFlowApplicantOnly() {
         testPage.clickElementById("none__checkbox-additionalIncome");
         testPage.clickContinue();
 
-        assertThat(testPage.getTitle()).isEqualTo(message("moneyonhand-types.title"));
-        testPage.clickElementById("none__checkbox-moneyOnHandTypes");
-        testPage.clickContinue();
+        //assertThat(testPage.getTitle()).isEqualTo(message("moneyonhand-types.title"));
+        //testPage.clickElementById("none__checkbox-moneyOnHandTypes");
+        //testPage.clickContinue();
```
spokenbird, line 664: Remove commented out code

src/main/java/org/mdbenefits/app/submission/conditions/HasMoreThan100OnHand.java
bseeger: This is a tough one, as I'm not sure which the default should be. Unset == truthy here, so it would look like they have more than 100 on hand. Thinking out loud here: 

If a default of "false" makes it default to truthy (HasMoreThan100OhHand = true), then the assumption is that they do have more than 100 on hand and they will get asked more about how much they have on hand.   

If a default to "true" makes it default to "falsey" (HasMoreThan100OnHand= false) then it looks like they have LESS than 100 on hand and we might skip the question of how much they have on hand - which may result in us not marking them as XSNAP.  

It seems like the first scenario is more in their favor -- assuming false and then the only extra is asking how much money on hand they have.  

Thoughts?

PR #257, opened 2024-05-22
Enable the session continuity interceptor (part of FFB).

bseeger:
Users will get redirected to the index page in the event of a session timeout.  This will prevent users from continuing on w/o knowing that their session expired and their record is now split.

Remove creation of a session in static page controller, as this is not necessary.


bseeger:
@spokenbird here's a pass at this. I think we should skip this for the current release, but try for the next one. There are a few things to be worked out as I don't think this is quite right yet. And I think the ffl needs a change as well to address an issue in the session interceptor code.  At any rate, this is a first pass, but I don't feel confident moving forward with this as I think it will still throw an error when a session times out.

bseeger:
Screenshot 2024-05-23 at 10 28 34 AM

PR #259, opened 2024-05-23
Fix expense calculator bug

spokenbird:


bseeger (approved):
LGTM. Thanks for the quick fix, @spokenbird

src/main/java/org/mdbenefits/app/utils/ExpenseCalculator.java
```diff
@@ -26,11 +26,12 @@ public BigDecimal totalUtilitiesExpenses() {
         Map<String, Object> inputData = submission.getInputData();
 
         expenses.forEach(val -> {
-            String inputFieldName = HomeExpensesType.getEnumByName(val).getInputFieldName();
-            expenseAmounts.add(
-                    new BigDecimal(inputData.getOrDefault(inputFieldName, "0").toString())
-                            .setScale(2, RoundingMode.HALF_UP));
-
+            if (!val.equalsIgnoreCase("None")) {
```
bseeger, line 29: @spokenbird I changed this to "equalsIgnoreCase()" so that we still do the right thing if someone accidentally changes this to "NONE" or "none".

src/main/resources/flows-config.yaml
```diff
@@ -82,6 +82,7 @@ flow:
     nextScreens:
       - name: expeditedSnapMoneyOnHand
   expeditedSnapMoneyOnHand:
+    onPostAction: MaybeClearMoneyOnHandAmount
     beforeSaveAction: CheckExpeditedSnapEligibility
```
bseeger, line 86: I added a new action to clear out the money amount in the event that they said they had money, but then went back and changed their mind on that.

PR #274, opened 2024-06-06
Fix expense calculator to handle blank entries for home expense amounts

bseeger:
Addresses: https://www.pivotaltracker.com/story/show/187749888

### DIFF
--- a/src/main/java/org/mdbenefits/app/utils/ExpenseCalculator.java
+++ b/src/main/java/org/mdbenefits/app/utils/ExpenseCalculator.java
@@ -28,9 +28,12 @@ public class ExpenseCalculator {
         expenses.forEach(val -> {
             if (!val.equalsIgnoreCase("None")) {
                 String inputFieldName = HomeExpensesType.getEnumByName(val).getInputFieldName();
-                expenseAmounts.add(
-                        new BigDecimal(inputData.getOrDefault(inputFieldName, "0").toString())
-                                .setScale(2, RoundingMode.HALF_UP));
+                // expense values are not required and may be blank
+                String expenseValue = inputData.getOrDefault(inputFieldName, "0").toString();
+                if (expenseValue.isBlank()) {
+                    return;
+                }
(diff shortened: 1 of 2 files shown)


src/main/java/org/mdbenefits/app/utils/ExpenseCalculator.java
```diff
@@ -28,9 +28,12 @@ public BigDecimal totalUtilitiesExpenses() {
         expenses.forEach(val -> {
             if (!val.equalsIgnoreCase("None")) {
                 String inputFieldName = HomeExpensesType.getEnumByName(val).getInputFieldName();
-                expenseAmounts.add(
-                        new BigDecimal(inputData.getOrDefault(inputFieldName, "0").toString())
-                                .setScale(2, RoundingMode.HALF_UP));
+                // expense values are not required and may be blank
+                String expenseValue = inputData.getOrDefault(inputFieldName, "0").toString();
+                if (expenseValue.isBlank()) {
+                    return;
```
spokenbird, line 34: Should this not be `continue`? `return` will stop the loop all together and return control the caller right? Where `continue` will stop the current iteration and move to the next?

bseeger: I think that because it's an anonymous function, that we need `return` versus `continue`, since the function is called once per entry, rather than it being a standard loop.  I'll confirm this though.

bseeger: Lambda's need the `return`. Essentially a `return` equals `break` in this context.  The code won't even compile if I use `break` there, which makes sense as it's not really in a for loop, but rather a function call.  (if you think of the thing that calls this lambda as the "for loop" this makes a lot of sense).

spokenbird: That makes sense, in that case this looks good to go!

spokenbird (approved):

PR #280, opened 2024-06-10
Update ohep rent screen

spokenbird:
- Needs link for I dont know

### DIFF
--- a/src/main/resources/templates/mdBenefitsFlow/ohepRent.html
+++ b/src/main/resources/templates/mdBenefitsFlow/ohepRent.html
@@ -32,7 +32,7 @@
               <th:block th:replace="~{fragments/inputs/submitButton :: submitButton(
                 text=#{general.inputs.continue})}"/>
               <div>
-                <a href="#" class="link" th:text="#{general.i-dont-know}"></a>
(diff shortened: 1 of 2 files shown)


src/main/resources/templates/mdBenefitsFlow/ohepRent.html
```diff
@@ -30,6 +31,9 @@
             <div class="form-card__footer">
               <th:block th:replace="~{fragments/inputs/submitButton :: submitButton(
                 text=#{general.inputs.continue})}"/>
+              <div>
+                <a href="#" class="link" th:text="#{general.i-dont-know}"></a>
+              </div>
```
bseeger, line 36: The "I don't know" should just take them to the next page, I think.  That's what I did on the next two pages. Probably good to check in with Devon though.

bseeger: I asked in Slack to confirm.

bseeger: Oh, just noting that this is in the ticket:  ```Add subtle link "I don't know" to enable skipping screen``` so seems like it should be a simple skip.

bseeger (approved):
LGTM!

PR #283, opened 2024-06-11
Add env var to prod deploy (addresses annoying error level log message in google drive code)

bseeger:
Addresses: https://www.pivotaltracker.com/story/show/187348029


bseeger:
I added this env var directly in aptible and watched the message go away.

spokenbird (approved):

